PR #1, opened 2021-12-17
CVL Overview section

mdgeorge4153:
I added an overview to the syntax section, and fleshed out the syntactic conventions.  I also moved up the common elements.


Nick-Certora:
Everything in here looks good

PR #5, opened 2022-02-28
described magic links

mdgeorge4153:
I migrated the information about which links are available from the internal docs: statsdata.json, FinalResults.html, Results.txt, zip output, job status.


mdgeorge4153:
The generated docs are available here: https://certora-certora-prover-documentation--5.com.readthedocs.build/en/5/docs/ref-manual/portal/secrets.html

docs/ref-manual/portal/secrets.md
```diff
@@ -0,0 +1,134 @@
...
+
+```
+https://vaas-stg.certora.com/jobStatus/65540/270dce9623d492937d82/?anonymousKey=6edb93d7abae7871f1c3be3b10863d64d2d72fef
+                             ─────────
+```
+
+FinalResults.html and Results.txt
+---------------------------------
+
+The `FinalResults.html` is an older version of the verification report, but
```
shellygr, line 37: Maybe put a nice note box saying Certora is working on newer report formats that will make these ones obsolete

docs/ref-manual/portal/secrets.md
```diff
@@ -0,0 +1,134 @@
...
+analyses fail, which can lead to timeouts.
+
+The `statsdata.json` file contains a section called `ANALYSIS` that lists all
+of the analyses that are attempted, and whether they succeed or fail.  If you
+see that a particular method consistently causes timeouts, you can check the
+ANALYSIS section to see if that method was correctly analyzed.  The analysis
+section should have a `true` for each successful analysis, and a `false` for
+each unsuccessful analysis.
+
+For example, the following shows that the `UNPACKING` analysis is failing on
```
shellygr, line 81: notably, as @jtoman can tell, not always a FALSE in UNPACKING is indicating a bad thing.
So it's worth saying that the exact interpretation of data in statsdata.json is subject to discussing with Certora's support

docs/ref-manual/portal/secrets.md
```diff
@@ -0,0 +1,134 @@
...
+several fields:
+
+![example statsdata.json showing several unpacking failures][statsdata.png]
+
+If an analysis is failing and causing you timeouts, contact the Certora team.
```
shellygr, line 86: nobody knows if an analysis failure is a cause for timeouts though ;)
so maybe phrase this a bit differently.
"Some timeouts could be caused by underlying analyses failures, so contact the Certora team in case you notice such failures"

docs/ref-manual/portal/secrets.md
```diff
@@ -0,0 +1,134 @@
...
+This link will allow you to download a tar file.  The submitted specs are
+contained in the `TarName` directory, and the original contracts can be found
+in the `TarName/input/.certora_config` directory (the file names will be
+changed).
+
+
+Other information
+-----------------
+
+Certora developers can access additional links; see [the internal documentation](https://certora.atlassian.net/wiki/spaces/CER/pages/278593893/Useful+Links+Certora+API)
```
shellygr, line 133: I'm not sure  our public docs should link to internal docs. Unless you can make this part visible only if you're logged in with a certora.com address

docs/ref-manual/portal/secrets.md
```diff
@@ -0,0 +1,134 @@
...
+but they can be useful for users too in some cases.  This page lists these
+outputs and explains how to read them.
+
+
+Job Status
+----------
+
+The job status page contains information about the job, including the arguments
+that were passed to the command line, the job message, and the exact certora
+prover version.  It is created as soon as the job is submitted, and also shows
```
shellygr, line 16: Do we care about capitalization? "Prover" or "prover"?

shellygr (commented):
few comments but overall looks highly informative

docs/ref-manual/portal/secrets.md
mdgeorge4153: Do you think it's worth adding more specificity about what to expect (e.g. something like "pointsto should is expected to fail on constructors" or whatever), or should I just point them to a section of the forum where we can collect that kind of information?

docs/ref-manual/portal/secrets.md
mdgeorge4153: Fixed

docs/ref-manual/portal/secrets.md
mdgeorge4153: Yeah, I was on the fence but I think I agree with you on this; will fix.

docs/ref-manual/portal/secrets.md
shellygr: I think better just point to the forum rather than be specific (but maybe pointsto can still be mentioned, I don't see this failure disappearing anytime soon)

docs/ref-manual/portal/secrets.md
mdgeorge4153: Updated

shellygr (approved):

docs/ref-manual/portal/secrets.md
mdgeorge4153: John pointed out that the UNPACKING example isn't interesting; I changed the example to a POINTSTO failure.  I also added a note about uninteresting failures.

PR #7, opened 2022-03-07
User Defined Types Documentation

thomasbernardi:


docs/ref-manual/cvl/types.md
```diff
@@ -89,6 +92,94 @@ multi-dimensional arrays
 reference types
 ```
 
+Types Defined in Solidity Contracts
```
mdgeorge4153, line 95: The above is already a section about types taken from Solidity, so I wouldn't start a new top-level section.  I might move the subsections of this section (enums / structs / value types) to between arrays and methods.

docs/ref-manual/cvl/types.md
```diff
@@ -37,6 +37,9 @@ CVL currently supports the following [solidity types][]:
  * `address`
  * `string`, `bytes`, and the sized `bytes` variants (`bytes1` through `bytes32`)
  * Tuples and single-dimensional arrays (both statically- and dynamically-sized)
+ * Enum Types
```
mdgeorge4153, line 40: Capitalization: Enum types, similarly below.

docs/ref-manual/cvl/types.md
```diff
@@ -37,6 +37,9 @@ CVL currently supports the following [solidity types][]:
...
+ * User Defined Value Types
```
mdgeorge4153, line 41: User-defined (hyphenate when two words are used as a single adjective or adverb, IIRC)

docs/ref-manual/cvl/types.md
```diff
@@ -89,6 +92,94 @@ multi-dimensional arrays
...
 
+Types Defined in Solidity Contracts
+-----------------------------------
+
+Solidity supports several user defined types:
+* Enums
+* User Defined Value Types (type aliases)
+* Structs
+
+These types can be accessed by referencing the contract where they are defined. To reference a contract remember to include `using ContractName as myLocalContractName`. The file containing the contract `ContractName` will have to be included in your build.
```
mdgeorge4153, line 103: We should add `id.id` to the syntax above.

mdgeorge4153, line 103: I would include less detail here, and instead point to the page on `using`.  I think this subsection could be condensed to a sentence, which could then just be included in the three sections below.  See suggestions under Enum

docs/ref-manual/cvl/types.md
```diff
@@ -89,6 +92,94 @@ multi-dimensional arrays
...
+
+User defined types may then be referenced from that contract: `myLocalContractName.myTypeName`. For types declared at the file level, any contract which imports that file may be used. Note that new user defined types may *not* be declared within a specification.
+
+### Enum Types
+Consider the following enum declaration and spec preamble:
+
+```
```
mdgeorge4153, line 110: specify `cvl` or `solidity` after the triple backtick.  See `README.md` in the top-level directory for an example

docs/ref-manual/cvl/types.md
```diff
@@ -89,6 +92,94 @@ multi-dimensional arrays
...
+contract TheContract {
+    enum MyEnum {
+        member1,
+        member2,
+        member3
+    }
+}
+...
+
+using TheContract as c
```
mdgeorge4153, line 120: Don't mix CVL and Solidity - make separate blocks and label them

docs/ref-manual/cvl/types.md
```diff
@@ -89,6 +92,94 @@ multi-dimensional arrays
...
+```
+
+The following are available in the entire spec:
+* The Type: `c.MyStruct`
+* For an expression `exp` of type `c.MyStruct` the following:
+    * `exp.member1`
+    * `exp.member2`
+    * `exp.member3`
+
+CVL Limitations: Struct assignments are allowed only in very limited situations. Otherwise, structs should be declared and then constrained, for example:
```
mdgeorge4153, line 174: What are the situations?

docs/ref-manual/cvl/types.md
```diff
@@ -89,6 +92,94 @@ multi-dimensional arrays
...
+        member2,
+        member3
+    }
+}
+...
+
+using TheContract as c
+```
+
+The following are available in the entire spec:
```
mdgeorge4153, line 123: You mention "in the entire spec" in several places; I think it's only worth mentioning if there are places where things are _not_ available.

docs/ref-manual/cvl/types.md
```diff
@@ -89,6 +92,94 @@ multi-dimensional arrays
...
+    * `c.MyEnum.member2`
+    * `c.MyEnum.member3`
+
+
+
+### User Defined Value Types
+
+Consider the following user defined value type declaration and spec preamble:
+
+```
```
mdgeorge4153, line 136: add solidity and CVL as  above

docs/ref-manual/cvl/types.md
```diff
@@ -89,6 +92,94 @@ multi-dimensional arrays
...
+* The Type: `c.MyEnum`
+* The values:
+    * `c.MyEnum.member1`
+    * `c.MyEnum.member2`
+    * `c.MyEnum.member3`
+
+
+
+### User Defined Value Types
+
```
mdgeorge4153, line 133: I might combine this with the previous section and move them all above.  Something like "Enums and type aliases defined in Solidity can be referenced from CVL."

Is this syntax different from Solidity?  I would assume the reader knows the Solidity syntax already and only mention differences.

mdgeorge4153 (changes requested):
This looks good.  I made a lot of picky comments.  Primarily, I think that for the reference manual all of this information should be condensed considerably, especially in places where the syntax is the same as Solidity's.

Let me know if you want to make an editing pass, some of my comments may be unclear or self-contradictory =)

docs/ref-manual/cvl/types.md
```diff
@@ -37,6 +37,9 @@ CVL currently supports the following [solidity types][]:
  * `address`
  * `string`, `bytes`, and the sized `bytes` variants (`bytes1` through `bytes32`)
  * Tuples and single-dimensional arrays (both statically- and dynamically-sized)
+ * Enum Types
+ * User Defined Value Types
+ * Struct Types
 
 The following are unsupported:
  * enums
```
mdgeorge4153, line 45: now-supported features should be removed from unsupported feature list

docs/ref-manual/cvl/types.md
mdgeorge4153, line 103: Just out of curiousity, do types defined in `currentContract` need to be fully qualified?

mdgeorge4153 (commented):
A few more comments

docs/ref-manual/cvl/types.md
thomasbernardi: so where do I put the bit about importing contracts to access this class of types?

mdgeorge4153 (approved):
Updated post-conversation

PR #8, opened 2022-03-08
update for multidim ghost mappings 

alexandernutz:
The multidim ghost mappings PR has landed on master -- updated docs accordingly.

NB, this update is quite limited in scope to the "syntax update -- ghost mapping" page.
We might consider updating the whole ghost documentation to be based around the mapping-style syntax rather than the function-style syntax.

(also, tentatively fixes to links in README.md)


mdgeorge4153 (approved):
Looks good, thanks

PR #12, opened 2022-03-25
Progress filling in reference manual holes

mdgeorge4153:
I am working my way through the reference manual, trying to make sure that the outline is comprehensive and updating with pointers to the old documentation where it exists, and an explicit note (containing the word "undocumented" for easy grepping) for documentation that needs to be written.

This PR contains my progress through the "Expressions" page.  As part of making that page comprehensive, I manually summarized the grammar describing expressions, and made sure that there was a section pertaining to each part.  I would especially like reviewers to focus on that page.


docs/ref-manual/cvl/expr.md
```diff
@@ -1,85 +1,289 @@
...
+           | "msg"   "." [ "address" | "sender" | "value" ]
+           | "tx"    "." [ "origin" ]
+           | "length"
+           | "selector" | "isPure" | "isView" | "numberOfArguments" | "isFallback"
+
+special_vars ::=
+           | "lastReverted" | "lastHasThrown"
+           | "lastStorage"
+           | "allContracts"
+           | "lastMsgSig"
```
nd-certora, line 48: @shellygr allContracts and lastMsgSig, what are they and do we want to support them? maybe better remove from the docs

docs/ref-manual/cvl/expr.md
```diff
@@ -1,85 +1,289 @@
...
+
+Several of the CVL types have special fields; see {doc}`types` (particularly
+{ref}`env`, {ref}`method-type`, and {ref}`arrays`).
+
+There are also several built-in variables:
+
+ * `bool lastReverted` and `bool lastHasThrown` are boolean values that indicate whether
+   the most recent contract function reverted or threw an exception.  They are
+   only meaningful if the most recent contract function was called with `@withrevert`
+   (see {ref}`call-expr`).
```
nd-certora, line 228: maybe explain that they can be used after a call with no revert but will always be false. It's a common mistake:
f@withrevert(e,arg)
gomeGetter() => lastReverted

nd-certora (approved):
a few comments that can be moved to "todo"s

docs/ref-manual/cvl/expr.md
shellygr: We can try to remove them, see if all tests pass, and then remove from docs too

docs/ref-manual/cvl/expr.md
mdgeorge4153: I changed the sentence and added a note about this.  We should consider changing this behavior; I added a CVL design doc

docs/ref-manual/cvl/expr.md
mdgeorge4153: They are currently marked as undocumented in the section of the document where they should be described; I think it makes sense to leave it that way until either someone figures out what they do or they are removed.

PR #17, opened 2022-04-12
Update options.md

EyalHochCertora:
Update the documentation of the rule_sanity flag.
Two main changes should be done:
1. Documentation for all of the 3 sanity checks we have: Reachability, Assert-Vacuity, Require-Redundancy.
2. Document the new way to use this flag, with the attached values which may be one of the following: `none`,` basic`, `advanced`.

### DIFF
--- a/docs/ref-manual/cli/options.md
+++ b/docs/ref-manual/cli/options.md
@@ -120,7 +120,7 @@ When you have a rule with multiple assertions:
 ### `--rule_sanity`
 
 **What does it do?**
-This mode will do some sanity checks for each rule, based on the attached value, which is allowed to be one of the following: none, basic, advanced.
+This mode will do some sanity checks for each rule, based on the attached value, which is allowed to be one of the following: `none`, `basic`, `advanced`.
 There are 3 kinds of sanity checks:
 1. Reachability- checks that even when ignoring all the user-provided assertions, the end of the rule is reachable. Namely, that the combination of requirements does not create an “empty” rule that is always true.
 
--- a/docs/ref-manual/cli/options.md
+++ b/docs/ref-manual/cli/options.md
@@ -131,6 +131,7 @@ When you have a rule with multiple assertions:
 **What does it do?**
 This option enables sanity checking for rules.  The `--rule_sanity` option must
 be followed by one of `none`, `basic`, or `advanced`; these are described below.
+See {doc}`../checking/sanity` for more information about sanity checks.
 
 There are 3 kinds of sanity checks:
 
--- a/docs/ref-manual/cli/options.md
+++ b/docs/ref-manual/cli/options.md
@@ -142,8 +142,8 @@ A `require` is considered to be redundant if it can be removed without affecting
 For example, each `require` with expression which is semanticlly equivalant to tautology, will be considered as redundant.
 
 The `rule_sanity` flag accepts one of the following values: `none`, `basic`, `advanced`, to control which sanity checks should be executed.
-The `none` keyword behaves the same as not mentioning the rule_sanity flag in the configuration at all. No sanity-checks will be executed.
-The `basic` keyword is intended for running only the reachability check for all the rules and the assert-vacuity check, but only for invariants.
+The `none` keyword behaves the same as not mentioning the `rule_sanity` flag in the configuration at all. No sanity-checks will be executed.
+The `basic` keyword is intended for running only the reachability check for all the rules and the `assert-vacuity` check, but only for invariants.
 Using the `advanced` keyword, all the sanity checks will be executed, for all the invariants/rules.
 It is recommended to start with the `basic` mode, since using the `advanced` mode might results in some false positive alarms.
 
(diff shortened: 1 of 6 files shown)


docs/ref-manual/cli/options.md
```diff
@@ -130,6 +132,15 @@ An example of an “empty” rule:
 
 _We expect all rules to fail this check._ The exception is the fallback function, which might pass.
 
+2. Assert-Vacuity- checks for each assert command in the rule, whether the assert is vacuously true, namely, even if all the previous preconditions are removed, the assert is always evaluated to true.
```
mdgeorge4153, line 135: I would split this into two sentences. "...vacuously true.  A rule is vacuously true if ..."

docs/ref-manual/cli/options.md
```diff
@@ -130,6 +132,15 @@ An example of an “empty” rule:
...
+
+3. Require-Redundancy- checks for each require command in the rule, whether the require is redundant, namely, we could remove it without affecting the satisfiability of the rule.
```
mdgeorge4153, line 137: Same here.  "...redundant.  A require is redundant if ..."

docs/ref-manual/cli/options.md
mdgeorge4153, line 135: Use code case for `assert`

mdgeorge4153, line 135: What do you mean by previous preconditions?  Do you mean `require`s?

docs/ref-manual/cli/options.md
```diff
@@ -130,6 +132,15 @@ An example of an “empty” rule:
...
+
+The rule_sanity flag accepts one of the following values: [none, basic, advanced], to control which sanity checks should be executed.
```
mdgeorge4153, line 139: Use code case for `rule_sanity` and for `none`, `basic`, and `advanced`.

mdgeorge4153 (changes requested):
Thanks for updating this.  I think at some point we will want to write a more extensive section of the user guide about sanity checks.  I left a few comments but overall this looks good.

docs/ref-manual/cli/options.md
```diff
@@ -119,8 +119,10 @@ When you have a rule with multiple assertions:
 (--rule_sanity)=
 ### `--rule_sanity`
 
-**What does it do?**  
-This mode will check for each rule that even when ignoring all the user-provided assertions, the end of the rule is reachable. Namely, that the combination of requirements does not create an “empty” rule that is always true.
+**What does it do?**
+This mode will do some sanity checks for each rule, based on the attached value, which is allowed to be one of the following: `none`, `basic`, `advanced`.
+There are 3 kinds of sanity checks:
+1. Reachability- checks that even when ignoring all the user-provided assertions, the end of the rule is reachable. Namely, that the combination of requirements does not create an “empty” rule that is always true.
```
mdgeorge4153, line 125: A few formatting comments:
 - the subsequent paragraphs of a numbered list entry need to be indented 4 spaces.  Otherwise, they are not considered part of the list, so the next numbered section restarts at 1 (in the [generated documentation](https://certora-certora-prover-documentation--17.com.readthedocs.build/en/17/docs/ref-manual/cli/options.html#rule-sanity) there are two bullet 1s).  See https://www.markdownguide.org/basic-syntax/#adding-elements-in-lists
 - you should use triple backticks followed by the name of the language for multi-line code blocks.  This will give proper indentation and syntax highlighting.  In this case, you want
````
```cvl
rule empty_rule() {
  ...
}
```
````

File: docs/ref-manual/cli/options.md (hunk 1 above)
Comment (mdgeorge4153) [reply docs/ref-manual/cli/options.md]: Maybe the best way to say this would be "An assertion is vacuously true if it evaluates to true on every example that reaches it, including those that would normally be ruled out by `require` statements."?

File: docs/ref-manual/cli/options.md [hunk 5]
```diff
@@ -130,12 +132,25 @@ An example of an “empty” rule:
 
 _We expect all rules to fail this check._ The exception is the fallback function, which might pass.
 
+2. Assert-Vacuity- checks for each `assert` command in the rule, whether the `assert` is vacuously true.
+An `assert` is considered to be vacuously true if even after all the previous preconditions are removed, the `assert` is always evaluated to true.
+For example, each `assert` with expression which is semanticly equivalant to tautology, will be considered as vacuosly true.
```
mdgeorge4153, line 137: spelling: semantically

docs/ref-manual/cli/options.md
```diff
@@ -130,12 +132,25 @@ An example of an “empty” rule:
...
+
+3. Require-Redundancy- checks for each `require` command in the rule, whether the `require` is redundant.
+A `require` is considered to be redundant if it can be removed without affecting the satisfiability of the rule.
+For example, each `require` with expression which is semanticly equivalant to tautology, will be considered as redundant.
```
mdgeorge4153, line 141: spelling: semantically

docs/ref-manual/cli/options.md
EyalHochCertora: > Maybe the best way to say this would be "An assertion is vacuously true if it evaluates to true on every example that reaches it, including those that would normally be ruled out by `require` statements."?

Yes, but also should refer to the if statements the assert is nested in.

mdgeorge4153 (approved):

PR #18, opened 2022-04-21
Michael/invariants

mdgeorge4153:
I completed the page on invariants, including notes on sources of unsoundness, preserved blocks, and filters.

### DIFF
--- a/docs/ref-manual/cvl/invariants.md
+++ b/docs/ref-manual/cvl/invariants.md
@@ -47,7 +47,8 @@ possible values of the parameters.
 While verifying an invariant, the Prover checks two things.  First, it checks
 that the invariant is established after the constructor.  Second, it checks
 that the invariant holds after the execution of any contract method, assuming
-that it held before the method was executed.
+that it held before the method was executed (if it does hold, we say the method
+*preserves* the invariant).
 
 If an invariant always holds at the beginning of every method call, it is
 always safe to assume that it is true.  The
@@ -104,9 +105,37 @@ used with care.
 Filters
 -------
 
-```{todo}
-This feature is currently undocumented.
+For performance reasons, you may want to avoid checking an invariant is
+preserved by a particular method or set of methods.  Invariant filters provide
+a method for skipping verification on a method-by-method basis.
+
+```{caution}
+Filtering out methods while checking invariants is {term}`unsound`.
 ```
+
+To filter out methods from an invariant, add a `filtered` block after the
+expression defining the invariant.  The body of the `filtered` block must
+contain a single filter of the form `var -> expr`, where `var` is a variable
+name, and `expr` is a boolean expression that may depend on `var`.
+
+If the expression evaluates to `false` with `var` replaced by a given method,
+the Prover will not check that the method preserves the invariant.  For example,
+the following invariant will not be checked on the `deposit(uint)`
+method:
+
+```cvl
+invariant balance_is_0(address a)
+    balanceOf(a) == 0
+    filtered {
+        f -> f.selector != deposit(uint).selector
+    }
+```
+
+In this example, when the variable `f` is bound to `deposit(uint)`, the
+expression `f.selector != deposit(uint).selector` evaluates to `false`, so the
+method will be skipped.
+
+See {ref}`method-type` for a list of the fields available on `method` objects.
 
 Preserved blocks
 ----------------
@@ -114,4 +143,9 @@ Preserved blocks
 ```{todo}
 This feature is currently undocumented.
 ```
+
+Writing an invariant as a rule
+------------------------------
+
+
 
--- a/docs/ref-manual/cvl/invariants.md
+++ b/docs/ref-manual/cvl/invariants.md
@@ -22,7 +22,7 @@ Syntax
 invariant ::= "invariant" id
               [ "(" params ")" ]
               expression
-              [ "filtered" "{" id "->" expression { "," id "->" expression } "}" ]
+              [ "filtered" "{" id "->" expression "}" ]
               [ "{" { preserved_block } "}" ]
 
 preserved_block ::= "preserved"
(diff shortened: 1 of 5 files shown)


mdgeorge4153:
The primary document that needs reviewing is here: https://certora-certora-prover-documentation--18.com.readthedocs.build/en/18/docs/ref-manual/cvl/index.html

docs/ref-manual/cvl/invariants.md
```diff
@@ -28,29 +35,254 @@ preserved_block ::= "preserved"
...
+a method for skipping verification on a method-by-method basis.
+
+```{caution}
+Filtering out methods while checking invariants is {term}`unsound`.
 ```
 
+To filter out methods from an invariant, add a `filtered` block after the
+expression defining the invariant.  The body of the `filtered` block must
+contain a single filter of the form `var -> expr`, where `var` is a variable
+name, and `expr` is a boolean expression that may depend on `var`.
```
orpistiner, line 137: I think that it should be explicitly written that `var` captures a method, and as a result, has all the attributes that of a method struct, e.g., selector, isView, etc.

docs/ref-manual/cvl/invariants.md
```diff
@@ -28,29 +35,254 @@ preserved_block ::= "preserved"
...
+
+If the expression evaluates to `false` with `var` replaced by a given method,
+the Prover will not check that the method preserves the invariant.  For example,
```
orpistiner, line 140: the Prover will not check that the method preserves the invariant _unless there is a preserved block that specifies that method's signature (see Preserved blocks section below)._

orpistiner (commented):
Added comments

CowardlyBattleCat (commented):
For those pages which include EBNF descriptions of syntax, we should possibly state that the form is EBNF and link to a new page where we can give a small primer on how to read EBNF. 

My first thought is that the page might: 
(a) describe how to interpret `::=`, `|`, `[` x `]`, and `{` y`}`  
(b) provide an example to show the relationship between an EBNF description and a syntactically correct statement in in CVL
(c) link to a couple external sites to provide context for what EBNF is (wikipedia and something more academic, perhaps)

docs/ref-manual/cvl/invariants.md
mdgeorge4153: Why does the presence of a preserved block override the filter?  That seems confusing to me

mdgeorge4153:
> 

I think this is a good idea.  I've submitted a separate PR to add links to the existing syntax page, and I think another PR expanding that page would be a good idea

docs/ref-manual/cvl/invariants.md
mdgeorge4153: I updated this

docs/ref-manual/cvl/invariants.md
mdgeorge4153: I think this is incorrect behavior, but have added a note to the docs

docs/ref-manual/cvl/invariants.md
```diff
@@ -28,29 +35,269 @@ preserved_block ::= "preserved"
...
-```{todo}
-This section is incomplete.  See [the user guide](/docs/user-guide/bank/index)
-for an overview of invariants.
+In CVL, an invariant is a property of the contract state that is expected to be
+true whenever a contract method is not currently executing.  This kind of
+invariant is sometimes called a "representation invariant".
+
+Each invariant has a name, possibly followed by a set of parameters, followed
+by a boolean expression.  We say the invariant *holds* if the expression
+evaluates to true in every reachable state of the contract, and for all
```
shellygr, line 44: the word "reachable" is a bit misleading here because the tool does not know what are the reachable states, so it checks for all states. It's up to the user to refine that set of states. I agree it's a bit of a fine point but maybe worth using a precise language here.

docs/ref-manual/cvl/invariants.md
```diff
@@ -28,29 +35,269 @@ preserved_block ::= "preserved"
...
+        f -> f.selector != deposit(uint).selector
+    }
+```
+
+In this example, when the variable `f` is bound to `deposit(uint)`, the
+expression `f.selector != deposit(uint).selector` evaluates to `false`, so the
+method will be skipped.
+
+```{note}
+If there is a {ref}`preserved block <preserved>` for a method it will be
```
shellygr, line 168: maybe say that if one complements a filtered expression with preserved blocks for the filtered methods, this is one way to handle the unsoundness originally introduced by the filtered expression.

shellygr: also wondering whether the discussion of preserved blocks shouldn't precede the filtered expressions - but it's fine either way for me

docs/ref-manual/cvl/invariants.md
```diff
@@ -28,29 +35,269 @@ preserved_block ::= "preserved"
...
+
+```cvl
+invariant zero_address_has_no_balance_v2(env e)
+    balanceOf(e, 0) == 0
+    { preserved { require e.msg.sender != 0; } }
+```
+
+In this example, we require the `msg.sender` argument to `balanceOf` to be
+nonzero, but makes no restrictions on the environment for the call to the method
+we are checking for preservation.
```
shellygr, line 232: This requires some more explanation. I would show an example code snippet where this confusion leads to the wrong result

shellygr (approved):
overall this is an excellent guide, I raised a few points which I trust Mike to address in some form before merging

docs/ref-manual/cvl/invariants.md
mdgeorge4153: Fixed

docs/ref-manual/cvl/invariants.md
mdgeorge4153: I expanded the example as suggested.

PR #21, opened 2022-05-16
Migrate and complete methods block documentation

mdgeorge4153:
This PR rewrites most of the existing documentation on the methods block and summarization and merges it into the reference manual.


mdgeorge4153:
Generated documentation is here: https://certora-certora-prover-documentation--21.com.readthedocs.build/en/21/docs/ref-manual/cvl/methods.html

docs/ref-manual/cvl/methods.md
```diff
@@ -46,44 +55,280 @@ method_summary   ::= "ALWAYS" "(" value ")"
...
-                   | [ "with" "(" cvl_params ")" ] expression
+                   | [ "with" "(" "env" id ")" ] block
+                   | [ "with" "(" "env" id ")" ] expression
+                   | [ "with" "(" "env" id ")" ] id "(" [ id { "," id } ] ")"
 
-cvl_param ::= cvl_type [ id ]
+```
 
+```{todo}
+The `with(env id) block` and `with(env id) expression` productions come from
```
thomasbernardi, line 65: The only valid expression summaries are applications of ghosts and cvl functions (I do not recall if we allow definitions but I suspect not)

docs/ref-manual/cvl/methods.md
```diff
@@ -46,44 +55,280 @@ method_summary   ::= "ALWAYS" "(" value ")"
...
+in the contract being verified does not depend on any of the environment
+variables.  The results of this check are displayed on the verification report
+as a separate rule called `envfreeFuncsStaticCheck`.
+
 ```{todo}
-This document is incomplete.  See {doc}`/docs/confluence/advanced/methods-overview`,
-{doc}`/docs/confluence/advanced/summaries`, {doc}`/docs/confluence/advanced/internal-summaries`,
-and {doc}`/docs/confluence/advanced/expressive-summaries` for the old documentation
-about the methods block.
+There is a separate check called `envfreeFuncsAreNonpayable`.  Why is this
```
thomasbernardi, line 105: I always assumed the "static check" didn't require a call to z3 while the other envfree rules did?

docs/ref-manual/cvl/methods.md
```diff
@@ -46,44 +55,280 @@ method_summary   ::= "ALWAYS" "(" value ")"
...
-### Application policies (UNRESOLVED or ALL)
+ * If the code for the function is not known at verification time, then the
+   function call must be summarized.  If no summary is given, the default summary
+   type is {ref}`AUTO <auto-summary>`, whose behavior is determined by the type of
+   function call.  In this case, the verification report will contain a contract
+   call resolution warning.
 
-### `ALWAYS`, `CONSTANT`, `PER_CALLEE_CONSTANT`
+```{todo}
+The default application policy is currently undocumented.
```
thomasbernardi, line 152: The default policy is `ALL` with the exception of a dispatcher summary whose default policy is `UNRESOLVED`.

docs/ref-manual/cvl/methods.md
```diff
@@ -46,44 +55,280 @@ method_summary   ::= "ALWAYS" "(" value ")"
...
+```
+
+```{todo}
+The old documentation is ambiguous about the behavior of `UNRESOLVED` summaries
```
thomasbernardi, line 156: It is impossible to have an "unresolved" internal method, so internal methods will only be summarized if the resolution type is `ALL` (since the code for the function is known at verification time).

That being said, obviously if the internal call is part of an external call that was not resolved, we will know nothing about the internal function call (but the external call will already be summarized)

docs/ref-manual/cvl/methods.md
```diff
@@ -46,44 +55,280 @@ method_summary   ::= "ALWAYS" "(" value ")"
...
+
+ * The `NONDET` approximation makes no assumptions about the return values; each
+   call to the summarized method may return a different result.
+
+```{todo}
+The old documentation seems to imply that the ETH balances for calls to
+`ALWAYS`, `CONSTANT`, and `PER_CALLEE_CONSTANT` do not account for the ETH value
+passed in the function call, while `NONDET` does.
+
+This seems like the wrong behavior?
```
thomasbernardi, line 193: I mean, aren't all of the summaries "wrong" behavior?

But more seriously, what would it even mean to "account for the ETH value passed in the function call" even mean when you are replacing a function call with a either a pure function which returns a static value, or a havoc'd value (ALWAYS and CONSTANT) or a havoc'd value per callee (PER_CALLEE_CONSTANT). Like, these are unsound summarizations which just replace the entire function and know _nothing_ about the parameters or state or anything? Also, how does NONDET account for the ETH value??

docs/ref-manual/cvl/methods.md
```diff
@@ -46,44 +55,280 @@ method_summary   ::= "ALWAYS" "(" value ")"
...
+```
+
+```{todo}
+The following note from the old documentation needs clarification:
```
thomasbernardi, line 197: @jtoman ?

docs/ref-manual/cvl/methods.md
```diff
@@ -46,44 +55,280 @@ method_summary   ::= "ALWAYS" "(" value ")"
...
+ * Normal calls and constructors use the `HAVOC_ECF` approximation: they are
+   assumed to change the state of external contracts arbitrarily but to leave
+   the caller's state unchanged.
+
+ * Calls to library methods are assumed to change the caller's state in an
+   arbitrary way, but are assumed to leave the state of other contracts
+   unchanged.
+
+```{todo}
+The effect of library calls on the current contract's ETH balance is unclear.
```
thomasbernardi, line 290: @jtoman

docs/ref-manual/cvl/methods.md
```diff
@@ -46,44 +55,280 @@ method_summary   ::= "ALWAYS" "(" value ")"
...
+variables are not supported.
+
+Functions used as summaries are not allowed to call contract functions.  They
+may only accept parameter types that are expressible in solidity; extended CVL
+types like `method` and `mathint` cannot be passed as arguments.
+
+### Undocumented summaries
+
+```{todo}
+Block summaries are undocumented.  Expression summaries other than function
```
thomasbernardi, line 330: block summaries don't exist
expression summaries other than function summaries don't exist
don't let the parser fool you :P

docs/ref-manual/cvl/methods.md
thomasbernardi: additionally, `with (env e)` does nothing and should not be used in summaries

docs/ref-manual/cvl/methods.md
mdgeorge4153: The question is why do we need a check for nonpayability?

docs/ref-manual/cvl/methods.md
jtoman: Yeah, this is pretty in the weeds, but the amount of data returned by the caller does not necessarily need to honor the request of the caller (in fact, in the case of dynamically sized return payloads, there is no "up front" size that the caller can declare for the expected size of the return buffer).

Instead, there is a sequence of sanity checks in the caller to establish that the callee returned enough data to satisfy the expected return values. So, if contract A calls contract B expecting contract B to return two uint256 values, and contract B only returns 32 bytes, then the call within A fails.

In other words, too-small return buffers is a failure mode of calls that are completely ignored by the summaries: from the perspective of the caller, the call external is summarized in such a way that the call "appears" to have exactly returned the amount of data required.

The optimistic return size is a flag to avoid insufficient returndata reverts (presumably when running with `@withrevert`). Like the summaries above, with the optimisticReturnsize flag, any unresolved (havocing) function will still be assumed to return the necessary number of bytes (even if those bytes are completely arbitrary).

docs/ref-manual/cvl/methods.md
jtoman: external library calls are delegate calls, and by the yellow paper they cannot (and should not) transfer balances. Internal library calls are not message calls, and so do not directly change balances (unless the body of these functions have external, balance transferring calls)

docs/ref-manual/cvl/methods.md
```diff
@@ -46,44 +55,280 @@ method_summary   ::= "ALWAYS" "(" value ")"
...
+   unchanged.
+
+```{todo}
+The effect of library calls on the current contract's ETH balance is unclear.
+
+The following comment from the old docs should be clarified:
+
+If the contract you are verifying relies heavily on modification of ETH
+balances, it's recommended to identify the balance-modifying functions and mark
+them `HAVOC_ALL` if necessary.
```
jtoman, line 296: This is pretty weird advice, I have no idea

jtoman (commented):
sorry for the late turn around

docs/ref-manual/cvl/methods.md
mdgeorge4153: ok, I removed expression and blocks and with env, leaving only the function call syntax.  The section describing function summaries only lists CVL functions and ghosts as possiblities.

docs/ref-manual/cvl/methods.md
mdgeorge4153: fixed.

docs/ref-manual/cvl/methods.md
mdgeorge4153: fixed.

docs/ref-manual/cvl/methods.md
mdgeorge4153: I haven't run any tests, but I think this behavior is what I would expect so I will document it so and consider deviations from it to be bugs.

docs/ref-manual/cvl/methods.md
mdgeorge4153: I replaced the third bullet by "Calls to library methods and `delegatecall`s are assumed to change the caller's storage in an arbitrary way, but are assumed to leave ETH balances and the storage of other contracts unchanged."  Is this accurate now?

docs/ref-manual/cvl/methods.md
mdgeorge4153: removed

docs/ref-manual/cvl/methods.md
mdgeorge4153: I understand now - payable functions depend on the msg.value which is part of the environment.  Updated.

docs/ref-manual/cvl/methods.md
mdgeorge4153: I think I have accurately described the situation.

docs/ref-manual/cvl/methods.md
mdgeorge4153: I think I've accurately updated the documentation

docs/confluence/advanced/methods.md
```diff
@@ -1,11 +1,228 @@
...
+
+Internal Function Summaries
+===========================
+
+Summaries Always Inserted
+-------------------------
+
+Summaries for **external** functions are only inserted when an implementation
+of that function cannot be found, and so we default to some summary that we,
+the verifier, sees fit. _However_, internal functions can always be found and
```
CowardlyBattleCat, line 24: For clarity and grammar, I would change the beginning of line 24 to "...as the verifier, see fit."

docs/confluence/advanced/methods.md
```diff
@@ -1,11 +1,228 @@
...
+```cvl
+methods {
+  mul(uint256 x, uint256 y) => easier_multiplication(x, y);
+}
+```
+
+Note this specific summarization is very dangerous and may cause vacuity bugs.
+
+In simple cases, these summaries may be used to replace harnesses, though the
+fact that they cannot call contract functions limits the types of harnesses
```
jtoman, line 227: ?? summaries?

docs/ref-manual/cvl/methods.md
```diff
@@ -46,44 +55,251 @@ method_summary   ::= "ALWAYS" "(" value ")"
...
+
+The application policy determines which function calls are replaced by
+approximations.  See {ref}`summaries` for details.
+
+(summaries)=
+Which function calls are summarized
+-----------------------------------
+
+Method summaries apply to all calls with the matching ABI signature, including
+internal methods and external methods on all contracts.  There is currently no
```
jtoman, line 125: nit: internal methods don't have an ABI signature, and in point of fact, the signatures that we write in the methods block aren't (necessarily) ABI signatures.

docs/ref-manual/cvl/methods.md
```diff
@@ -46,44 +55,251 @@ method_summary   ::= "ALWAYS" "(" value ")"
...
+way to apply different summaries to different contracts or to summarize some
+calls and not others to methods with the same ABI signature.  For this reason,
```
jtoman, line 127: this is not, strictly speaking, true. It is possible that for some function, e.g., `owner()`, the callee for some call-sites is known and for others it is not. If you don't specify an forced summary, then the summary behavior will only apply for those callsites where the callee is unresolved.

docs/ref-manual/cvl/methods.md
```diff
@@ -46,44 +55,251 @@ method_summary   ::= "ALWAYS" "(" value ")"
...
+
+The Solidity compiler allows internal functions to have richer type
+specifications than just ABI specifications, but the prover only considers the
+ABI encoding to determine whether a method call should be summarized.  For
+example, internal functions can accept struct arguments, but the structs are
+replaced by tuples in the ABI signature.  If the generated ABI signature of an
+internal method call matches the ABI signature specified in the methods block,
+it will be a candidate for summarization even if the declared argument types in
+the Solidity function don't exactly match the declared argument types in the
+`methods` block.
```
jtoman, line 139: @thomasbernardi I thought this was no longer the case

docs/ref-manual/cvl/methods.md
```diff
@@ -46,44 +55,251 @@ method_summary   ::= "ALWAYS" "(" value ")"
...
+ * With `DISPATCHER(false)` or just `DISPATCHER`, the unknown contract is
+   considered as a possiblity
+
+ * With `DISPATCHER(true)`, only the known contract instances are considered
+
+The most commonly used option is `DISPATCHER(true)`, because in most cases the
+behavior of `DISPATCHER(false)` is equivalent to that of `AUTO`.
+`DISPATCHER(false)` only behaves differently from `AUTO` when one of the known
+contracts exhibits behavior that `AUTO` assumes is impossible (such as
+reentrancy).
```
jtoman, line 262: I don't think this is true? If the callee address is *always* one of the known contracts (as proven by z3) then the two dispatcher flavors are equivalent? Maybe you're making a point about possible static behaviors...

docs/confluence/advanced/methods.md
mdgeorge4153: same here

docs/ref-manual/cvl/methods.md
mdgeorge4153: my understanding was that we take the signature of the internal method or the methods block, erase all of the information that goes beyond what would be encoded in an ABI signature, and then use that for dispatching.  Is that incorrect?  It sounds like either way this needs clarification

docs/ref-manual/cvl/methods.md
mdgeorge4153: I think this behavior is described below, but I should be more careful here to distinguish between which methods match and which methods are approximated.

docs/ref-manual/cvl/methods.md
mdgeorge4153: We should discuss and maybe look at tests, because this was my understanding of what Thomas explained to me

docs/ref-manual/cvl/methods.md
thomasbernardi: > this is not, strictly speaking, true. It is possible that for some function, e.g., owner(), the callee for some call-sites is known and for others it is not. If you don't specify an forced summary, then the summary behavior will only apply for those callsites where the callee is unresolved.

I think this is not quite what we're talking about here. Perhaps there should be a more clear distinction in language between how summaries are "specified" and how the tool "applies" them.

docs/ref-manual/cvl/methods.md
thomasbernardi: Internal method's _shouldn't_ have an ABI signature and as far as `solc` is concerned they do not. However, as far as our tool is concerned, all methods have an ABI signature. Maybe we can use some language like "the equivalent ABI signature" or isomorphic or something.

mdgeorge4153: I think the latest draft clarifies this

docs/ref-manual/cvl/methods.md
mdgeorge4153: I think the latest draft clarifies this

docs/ref-manual/cvl/methods.md
mdgeorge4153: resolved in latest draft

docs/ref-manual/cvl/methods.md
mdgeorge4153: I weakened the claim in the current draft to something that I now think is true.

nd-certora (approved):

docs/ref-manual/cvl/methods.md
```diff
@@ -46,44 +55,251 @@ method_summary   ::= "ALWAYS" "(" value ")"
                    | "HAVOC_ALL"
                    | "DISPATCHER" [ "(" ( "true" | "false" ) ")" ]
                    | "AUTO"
-                   | [ "with" "(" cvl_params ")" ] block
-                   | [ "with" "(" cvl_params ")" ] expression
-
-cvl_param ::= cvl_type [ id ]
+                   | id "(" [ id { "," id } ] ")"
 
 ```
```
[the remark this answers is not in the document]
mdgeorge4153: I think this is a good idea except that it will lead to a lot of duplication.  This fact is included in the description below though.

docs/ref-manual/cvl/methods.md
```diff
@@ -1,21 +1,30 @@
...
+* **Summary declarations** are used to replace calls to methods having the
   given signature with something that is simpler for the Prover to reason about.
   Summaries allow the Prover to reason about external contracts whose code is
   unavailable.  They can also be useful to simplify the code being verified to
   circumvent timeouts.
 
+```{caution}
+Summary declarations change the way that some function calls are interpreted,
+and are therefore {term}`unsound` (with the exception of `HAVOC_ALL` summaries,
+which are always sound).
```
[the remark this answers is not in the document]
mdgeorge4153: fixed

docs/ref-manual/cvl/methods.md
```diff
@@ -46,44 +55,251 @@ method_summary   ::= "ALWAYS" "(" value ")"
...
+
+ * {ref}`const-summary`
+ * {ref}`havoc-summary`
+ * {ref}`dispatcher`
+ * {ref}`auto-summary`
+ * {ref}`function-summary`
+
+The application policy determines which function calls are replaced by
+approximations.  See {ref}`summaries` for details.
+
```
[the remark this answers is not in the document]
mdgeorge4153: @thomasbernardi ?

docs/confluence/advanced/methods.md
```diff
@@ -1,11 +1,228 @@
...
+following summaries are allowed:
+
+*   `ALWAYS(X)` the summary always returns `X` and has no side-effects
+    
+*   `CONSTANT` the summary always returns the same constant and has no side
+    effects
+    
+*   `NONDET` the summary returns a havoc'd value
+    
+*   `Ghost` the summary returns the value return by the given ghost function
```
thomasbernardi, line 50: CVL Function should be allowed too right?

mdgeorge4153:
We should update the pointer in the “Prover approximations” section to refer here.  We should also remove more of the old docs.

docs/confluence/advanced/methods.md
mdgeorge4153: confluence docs are deprecated.

docs/ref-manual/cvl/methods.md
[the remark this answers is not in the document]
mdgeorge4153: Thomas is designing a test suite to figure out exactly what the limitations are; in the mean time I have added the limitations  I know about and left a TODO that there are other unsupported combinations.  Once we have more accurate information I'll submit another PR.

PR #22, opened 2022-05-16
added support for spell checking

mdgeorge4153:


mdgeorge4153:
I wasn't able to get readthedocs to automatically check the spelling when we push, but I did make it automatically run when we build locally.  Fixed a bunch of typos.

docs/confluence/perplexed.md
```diff
@@ -43,12 +43,16 @@ SMT is the hardest nut to crack! There could be so many reasons that they occur.
...
+    ```
     then the file would be in  
+    ```
     https://vaas-stg.certora.com/output/43260/dafd64136c1d71ba882c/**statsdata.json**?anonymousKey=a8e7a2b10c4f9f74ea55624c166d7edee8f1e2bc  
+    ```
     Look at the `ANALYSIS` key. If you see any `false` entries on keys that are not `constructor()` or `ecrecover` - report in #bugs.
     
-2.  **A non-empty “Call Resolution” warnings table.** Delegatecalls could be tricky and invalidate some of the Prover’s optimizations that were built to avoid timeouts. If such appear, try to scope out the verification to ignore the delegatecalls. Check with a security person, a more experienced user, or even the customer if you’re not sure if it’s safe to ignore.
+2.  **A non-empty “Call Resolution” warnings table.** Delegate calls could be tricky and invalidate some of the Prover’s optimizations that were built to avoid timeouts. If such appear, try to scope out the verification to ignore the `delegatecall`s. Check with a security person, a more experienced user, or even the customer if you’re not sure if it’s safe to ignore.
```
CowardlyBattleCat, line 55: The first "Delegate calls" here now has a space. The second "`delegatecall`s does not have a space.

It seems that the first is written with the English syntax (the space is needed) while the second is written as a technical keyword (so the space should be omitted). Is this correct?

spelling_wordlist.txt
```diff
@@ -0,0 +1,87 @@
...
+conjuncted
+dereference
+dereferences
+enum
+enums
+envfree
+executable
+executables
+getter
+getter
```
CowardlyBattleCat, line 25: Should this second "getter" instead be "getters" given the pattern elsewhere in this list?

CowardlyBattleCat (approved):

docs/confluence/perplexed.md
mdgeorge4153: yes

spelling_wordlist.txt
mdgeorge4153: fixed

PR #24, opened 2022-05-26
User Defined Types

thomasbernardi:
I'm not really sure what the scope of these docs are so any requests on examples or explanations would be great.

Also, I wasn't sure where to put this file/how to link it so help with that would be great!


mdgeorge4153:
Have you looked at https://docs.certora.com/en/latest/docs/ref-manual/cvl/types.html?

mdgeorge4153:
In paarticular https://docs.certora.com/en/latest/docs/ref-manual/cvl/types.html

thomasbernardi:
I guess I'm confused what the distinction between the user guide and the reference docs is and whether I'm on the right track here.

PR #25, opened 2022-06-01
Invariants preserved blocks - adding support for a specific preserved block for the fallback() function.

GabrielE1966:
Currently there are two kinds of preserved blocks in an invariant definition. **Generic** - applied for all methods of a contract, and **Specific** - applied only for a specific method.  Adding, a third option,  a support for a preserved block specific for the `fallback()` function, that could not be implemented in the specific preserved block.  This will allow the users to add assumptions for the verification of the fallback function, the function that is executed in any call to the contract that doesn't match any of the contracts methods.

### DIFF
--- a/docs/ref-manual/cvl/invariants.md
+++ b/docs/ref-manual/cvl/invariants.md
@@ -29,10 +29,11 @@ invariant ::= "invariant" id
 
 preserved_block ::= "preserved"
                     [ method_signature ]
-                    [ "fallback()" ]
                     [ "with" "(" params ")" ]
                     block
 
+method_signature ::= id "(" [ evm_type [ id ] { "," evm_type [ id ] } ] ")"
+                     | "fallback" "(" ")"
 ```
 
 See {doc}`basics` for the `id` production, {doc}`expr` for the `expression`
@@ -155,15 +156,14 @@ commands are also possible.
 
 Preserved blocks are listed after the invariant expression (and after the filter
 block, if any), inside a set of curly braces (`{ ... }`).  Each preserved block
-consists of the keyword `preserved` followed by an optional method signature 
-(or `fallback()`), an optional `with` declaration, and finally the block of commands 
-to execute.
+consists of the keyword `preserved` followed by an optional method signature, 
+an optional `with` declaration, and finally the block of commands to execute.
 
-If a preserved block specifies a method signature, the signature should
+If a preserved block specifies a method signature, the signature either be `fallback()` or should
 match one of the contract methods, and the preserved block only applies when
-checking preservation of that contract method.  The fallback() preserved block
+checking preservation of that contract method.  The `fallback()` preserved block
 applies only to the `fallback()` function that should be defined in the contract.
-The arguments of the method arein scope within the preserved block.  
+The arguments of the method are in scope within the preserved block.  
 If there is no method signature, the preserved is a default block that applies to 
 all methods that don't have a specific preserved block, including the `fallback()` method.
 


docs/ref-manual/cvl/invariants.md
```diff
@@ -29,6 +29,7 @@ invariant ::= "invariant" id
 
 preserved_block ::= "preserved"
                     [ method_signature ]
+                    [ "fallback()" ]
```
mdgeorge4153, line 32: This should probably be "fallback" "(" ")".  Also, fallback should be part of `method_signature`; with the way it's written here you can write something like "preserved f(uint x) fallback () { ... }".  Better to add
```
method_signature ::= id "(" types ")"
     | "fallback" "(" ")"
```

mdgeorge4153: it should probably actually be
```
  id "(" [ evm_type [ id ] { "," evm_type [ id ] } ] ")"
| "fallback" "(" ")"
```

for the method_signature definition

docs/ref-manual/cvl/invariants.md
```diff
@@ -154,15 +155,17 @@ commands are also possible.
 
 Preserved blocks are listed after the invariant expression (and after the filter
 block, if any), inside a set of curly braces (`{ ... }`).  Each preserved block
-consists of the keyword `preserved` followed by an optional method signature,
-an optional `with` declaration, and finally the block of commands to execute.
+consists of the keyword `preserved` followed by an optional method signature 
+(or `fallback()`), an optional `with` declaration, and finally the block of commands
```
mdgeorge4153, line 159: if you change method_signature as described above, you can leave out the "or fallback" parenthetical, since it is a kind of method signature.

docs/ref-manual/cvl/invariants.md
```diff
@@ -154,15 +155,17 @@ commands are also possible.
 
 Preserved blocks are listed after the invariant expression (and after the filter
 block, if any), inside a set of curly braces (`{ ... }`).  Each preserved block
-consists of the keyword `preserved` followed by an optional method signature,
-an optional `with` declaration, and finally the block of commands to execute.
+consists of the keyword `preserved` followed by an optional method signature 
+(or `fallback()`), an optional `with` declaration, and finally the block of commands 
+to execute.
 
 If a preserved block specifies a method signature, the signature should
```
mdgeorge4153, line 162: ... should either be `fallback` or should match ...

docs/ref-manual/cvl/invariants.md
```diff
@@ -154,15 +155,17 @@ commands are also possible.
...
 match one of the contract methods, and the preserved block only applies when
-checking preservation of that contract method.  The arguments of the method are
-in scope within the preserved block.  If there is no method signature, the
-preserved is a default block that applies to all methods that don't have a
-specific preserved block.
+checking preservation of that contract method.  The fallback() preserved block
+applies only to the `fallback()` function that should be defined in the contract.
```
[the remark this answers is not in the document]
mdgeorge4153: if `fallback` applies to the default fallback as well, then I would say "applies to the contract's fallback function" and drop the "which should...".  Otherwise, I would probably make it clearer that it does not apply to the default fallback

mdgeorge4153 (changes requested):
Looks good, modulo some editorial comments left below.  If the syntax isn't finalized, I would consider making it `fallback` instead of `fallback()`, although I'm on the fence about which I prefer.

mdgeorge4153:
Is this an upcoming feature, or already implemented?  If it's upcoming, we should maybe wait to merge the documentation until the feature lands and/or we start versioning the documentation

docs/ref-manual/cvl/invariants.md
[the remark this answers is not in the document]
GabrielE1966: I didn't see any reference in the Solidity documentation to a "default fallback" function being created by the solc. What I understood is that in case Ether is sent to a contract without fallback (or `receive`) function, an exception is raised and the Ether is sent back to the caller.

docs/ref-manual/cvl/invariants.md
GabrielE1966: done

docs/ref-manual/cvl/invariants.md
GabrielE1966: done

docs/ref-manual/cvl/invariants.md
GabrielE1966: done

docs/ref-manual/cvl/invariants.md
```diff
@@ -154,15 +155,17 @@ commands are also possible.
...
+(or `fallback()`), an optional `with` declaration, and finally the block of commands 
+to execute.
 
 If a preserved block specifies a method signature, the signature should
 match one of the contract methods, and the preserved block only applies when
-checking preservation of that contract method.  The arguments of the method are
-in scope within the preserved block.  If there is no method signature, the
-preserved is a default block that applies to all methods that don't have a
-specific preserved block.
+checking preservation of that contract method.  The fallback() preserved block
```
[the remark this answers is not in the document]
GabrielE1966: done

docs/ref-manual/cvl/invariants.md
```diff
@@ -154,15 +155,17 @@ commands are also possible.
...
+applies only to the `fallback()` function that should be defined in the contract.
+The arguments of the method arein scope within the preserved block.
```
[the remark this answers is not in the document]
GabrielE1966: corrected

docs/ref-manual/cvl/invariants.md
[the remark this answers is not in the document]
mdgeorge4153: ok, sounds good.

mdgeorge4153 (approved):
Looks good

PR #28, opened 2022-06-28
document hashing approximation

alexandernutz:
An attempt to document the imprecision in our hashing of arrays like `bytes` etc. 
It's not perfect, but should work as some reference for when this comes up.

(we should work on better solutions on the tool side, too -- but that might take some time ...)

### DIFF
--- a/docs/ref-manual/approx/array-hashes.md
+++ b/docs/ref-manual/approx/array-hashes.md
@@ -1,0 +1,44 @@
+Array hashes
+============
+
+### Background: Non-Collision of Hashes is Essential for Storage Integrity
+
+Solidity makes intensive use of the Keccak hash function in order to create
+its storage layout.
+For instance consider this contract:
+```solidity
+contract A {
+  mapping(uint => uint) m; // first field of A -- gets slot 0
+
+  function setMTo0(uint i) {
+    m[i] = 0
+  }
+}
+
+```
+In order to determine the storage address of `m[i]`, solidity 
+computes this hash: `keccak(0, i)` (the first argument is 0 because mapping `m` 
+is the first field in contract `A`).
+
+This means that non-collision of hashes is crucial for memory integrity, since 
+a collision means that writes to different solidity variables interfere with 
+each other.
+
+### Modelling of Hashing in CVT -- Adequate, Except for Unbounded Arrays
+
+CVT generally uses a modelling of the Keccak function that prohibits these 
+unwanted collisions.
+An exception to this statement is the case when we hash arrays of unbounded 
+length, like `bytes`, or `string`.
+If such a `bytes` array is longer than the bound chosen via the 
+`--settings -byteMapHashingPrecision=X` CVT setting, CVT will be allowed to 
+choose an arbitrary hash value as a result.
+Note also that CVT may be allowed to choose the length of the array.
+
+This can easily lead to unexpected collisions.
+A workaround for avoiding these collisions is to introduce `requires` statements
+that bound the mapping to not exceed the bound for precise hashing.
+
+The dev team is looking into more user-friendly solutions to this problem.
+
+
--- a/docs/ref-manual/approx/index.md
+++ b/docs/ref-manual/approx/index.md
@@ -16,5 +16,6 @@ that they can cause real bugs to be overlooked by the Prover.
 loops.md
 summarization.md
 harnessing.md
+array-hashes.md
 ```
 


docs/ref-manual/approx/array-hashes.md
```diff
@@ -0,0 +1,44 @@
...
+computes this hash: `keccak(0, i)` (the first argument is 0 because mapping `m` 
+is the first field in contract `A`).
+
+This means that non-collision of hashes is crucial for memory integrity, since 
+a collision means that writes to different solidity variables interfere with 
+each other.
+
+### Modelling of Hashing in CVT -- Adequate, Except for Unbounded Arrays
+
+CVT generally uses a modelling of the Keccak function that prohibits these
```
mdgeorge4153, line 29: Does this approximation apply to other uses of keccak hashing?  If so, we should generalize this document and describe the set of places we make this assumption.

mdgeorge4153, line 29: modeling

docs/ref-manual/approx/array-hashes.md
```diff
@@ -0,0 +1,44 @@
...
+each other.
+
+### Modelling of Hashing in CVT -- Adequate, Except for Unbounded Arrays
+
+CVT generally uses a modelling of the Keccak function that prohibits these 
+unwanted collisions.
+An exception to this statement is the case when we hash arrays of unbounded 
+length, like `bytes`, or `string`.
+If such a `bytes` array is longer than the bound chosen via the 
+`--settings -byteMapHashingPrecision=X` CVT setting, CVT will be allowed to
```
mdgeorge4153, line 34: Link to the setting (it should be documented in `ref-manual/cli/options.md`)

docs/ref-manual/approx/array-hashes.md
```diff
@@ -0,0 +1,44 @@
...
+This means that non-collision of hashes is crucial for memory integrity, since 
+a collision means that writes to different solidity variables interfere with 
+each other.
+
+### Modelling of Hashing in CVT -- Adequate, Except for Unbounded Arrays
+
+CVT generally uses a modelling of the Keccak function that prohibits these 
+unwanted collisions.
+An exception to this statement is the case when we hash arrays of unbounded 
+length, like `bytes`, or `string`.
```
mdgeorge4153, line 32: I might say "including `bytes` and `string`", since the most common case of unbounded arrays is something like `int[]`.

docs/ref-manual/approx/array-hashes.md
```diff
@@ -0,0 +1,44 @@
...
+If such a `bytes` array is longer than the bound chosen via the 
+`--settings -byteMapHashingPrecision=X` CVT setting, CVT will be allowed to 
+choose an arbitrary hash value as a result.
+Note also that CVT may be allowed to choose the length of the array.
+
+This can easily lead to unexpected collisions.
+A workaround for avoiding these collisions is to introduce `requires` statements
+that bound the mapping to not exceed the bound for precise hashing.
+
+The dev team is looking into more user-friendly solutions to this problem.
```
mdgeorge4153, line 42: development

docs/ref-manual/approx/array-hashes.md
```diff
@@ -0,0 +1,44 @@
...
+```
+In order to determine the storage address of `m[i]`, solidity 
+computes this hash: `keccak(0, i)` (the first argument is 0 because mapping `m` 
+is the first field in contract `A`).
+
+This means that non-collision of hashes is crucial for memory integrity, since 
+a collision means that writes to different solidity variables interfere with 
+each other.
+
+### Modelling of Hashing in CVT -- Adequate, Except for Unbounded Arrays
```
mdgeorge4153, line 27: Modeling

docs/ref-manual/approx/array-hashes.md
```diff
@@ -0,0 +1,44 @@
+Array hashes
+============
+
+### Background: Non-Collision of Hashes is Essential for Storage Integrity
+
+Solidity makes intensive use of the Keccak hash function in order to create
```
mdgeorge4153, line 6: Add Keccak to `/spelling_wordlist.txt`.

mdgeorge4153: note that local builds run spell checking when you run `make`, so you can use this to spell check

docs/ref-manual/approx/array-hashes.md
```diff
@@ -0,0 +1,44 @@
+Array hashes
+============
+
+### Background: Non-Collision of Hashes is Essential for Storage Integrity
```
mdgeorge4153, line 4: This should be a level-2 header instead of level-3 (i.e. `##` instead of `###`).  Sphinx complains if you jump levels.

docs/ref-manual/approx/array-hashes.md
mdgeorge4153, line 27: This also needs to be `##` instead of `###`

docs/ref-manual/approx/array-hashes.md
```diff
@@ -0,0 +1,44 @@
+Array hashes
+============
+
```
mdgeorge4153, line 3: Somewhere up front we should clearly explain the approximation.  "The Prover does not model hashing directly.  Instead, it makes the following assumptions about `hash(x)`: - if `x != y` then `hash(x) != hash(y)`, etc.

We should also clearly explain the surprising thing the user might see.  A clear example of a rule that should pass and doesn't, or a rule that shouldn't pass and does, and an explanation of why.

docs/ref-manual/approx/array-hashes.md
mdgeorge4153, line 4: Is the storage integrity issue related to the approximation?  It's hard to tell without the example.  Does the hashing approximation mean that if I write to `m["some really long string"]` I just havoc everything because memory integrity is violated?

mdgeorge4153 (changes requested):
I think this needs some more detail to clearly explain the issue to the user, some suggestions inline.  Also, running `make` produces warnings.

alexandernutz:
I'm making a new branch/PR for this (`alex/hashing`) -- I'll carry over what's useful here, including trying to integrate the comments from the discussion of course...

PR #30, opened 2022-07-11
Mention solidity version constraint

kustosz:
This document was sent to me as "a small example to verify that everything is set up properly" and took me for a spin thinking my setup was indeed not proper. Should save future readers some friction.


mdgeorge4153 (approved):
Looks good, thanks!

PR #33, opened 2022-07-22
multicontract user guide

mdgeorge4153:
A user guide chapter on multicontract support.  The generated chapter is [here](https://certora-certora-prover-documentation--33.com.readthedocs.build/en/33/docs/user-guide/multicontract/index.html).


docs/user-guide/multicontract/index.md
```diff
@@ -0,0 +1,675 @@
...
+`transferFrom` to change the balances any way it likes.  In fact, calls to
+`asset.balanceOf(...)` are also unresolved, so the Prover can choose any return
+value that causes a counterexample.  In this case, we can see that the Prover
+chose `3` for the first return value of `balanceOf` and `9`
+for the last return value of `balanceOf`:
+
+![Call trace for `integrityOfDeposit` on `Pool` showing call to `assetBalance` with internal havoced call to `balanceOf`, returning 3 in once place and 9 in another](no-link-variables.png)
+
+The "Call Resolution" tab on the report provides more information about all of
+the unresolved method calls within the contract and how they are resolved by
```
nd-certora, line 163: would external calls be a better term? because you see both resolved and still unresolved.

docs/user-guide/multicontract/index.md
```diff
@@ -0,0 +1,675 @@
...
+```
+
+The `--link Pool:asset=Asset` option tells the Prover to assume that the `asset`
+field of the `Pool` contract instance in the scene is a pointer to the `Asset`
+contract instance in the scene.  With this information, the Prover is able to
+resolve the calls to the methods on `Pool.asset` using the code in `Asset.sol`.
+
+With this option, the Prover is no longer able to construct a counterexample to
+the `integrityOfDeposit` rule, so the rule passes.
+
```
nd-certora, line 218: should we say something about the call resolution - it does not apper there any more. right? ( very confusing what is in and is not in the call resolution) 
We can also give links to result page so one can see how it should look

mdgeorge4153:
Need to add a note that sometimes the linking gets lost and turns unresolved - need to ask @nd-certora

docs/user-guide/multicontract/index.md
```diff
@@ -0,0 +1,675 @@
...
+Working with known contracts
+----------------------------
+
+In the case of the `Pool` verification, we don't want the Prover to choose
+arbitrary behavior for the `Asset`, because we have the `Asset` code.  Instead,
+we would like the Prover to model the `asset` contract using the `Asset` code.
+
+To do so, we must first add the `Asset` contract to the set of contracts that
+the Prover knows about.  This set of contracts is called the {term}`scene`.  You
```
mdgeorge4153, line 185: We should add a note here (or elsewhere) about the `--address` option (maybe a footnote)

docs/user-guide/multicontract/index.md
```diff
@@ -0,0 +1,675 @@
...
+As with the `callbackChoice` variable, the Prover will consider every possible
+value for the `arbitraryUint`, `arbitraryAddress1`, and `arbitraryAddress2`
+fields, so this code has the effect of calling an arbitrary method on `pool`
+with arbitrary arguments.
+
+This approach still doesn't give perfect coverage.  For example, the
+`FlexibleReceiver` will only call one contract method, while a real attack may
+require two or more calls.  If the contract calls the `executeOperation` more
+than once it will always have the same effect (although this can be changed by
+making the arbitrary fields into mappings).
```
0xDatapunk, line 605: the fuller example can use a counter as a mapping key to get arbitrary values

docs/user-guide/multicontract/index.md
mdgeorge4153: Updated.  I considered adding run links, but I think I asked and it wasn't clear if these are persistent enough to include in the documentation.  Even if we do, I would want to have more process in place to make sure we are rebuilding the examples when they change.

docs/user-guide/multicontract/index.md
mdgeorge4153: fixed

nd-certora (approved):
a few final comments

docs/user-guide/multicontract/index.md
```diff
@@ -0,0 +1,743 @@
...
+```
+
+[^addressOption]: You can control the address chosen for the contract instance
+using the {ref}`--address` option.
+
+Adding `Asset.sol` to the scene makes the Prover aware of it, but it does not
+connect the `asset` field of the pool to the `Asset` contract.  Although
+`Pool.asset` is declared to have type `Asset` in the solidity source, the
+solidity compiler erases that information from the bytecode; in the compiled
+bytecode the field is just treated as an `address`.
```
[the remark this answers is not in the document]
mdgeorge4153: I don't think this is true, in fact it could point to any contract at all (or even a non-contract).  I added a sentence though.

docs/user-guide/multicontract/index.md
```diff
@@ -0,0 +1,743 @@
...
+
+To reconnect the `Asset` code to the `Pool.asset` field, we can use the
```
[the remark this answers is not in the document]
mdgeorge4153: I settled on "connect"

docs/user-guide/multicontract/index.md
```diff
@@ -0,0 +1,743 @@
...
+   For example, a pool might be designed to work with arbitrary ERC20
+   implementations.
+
+In this case, the only option is to {term}`summarize` the unknown code for the
+Prover.  Although there are many available types of summaries, the ones most
+commonly used for unknown code are {ref}`dispatcher`.
+
+The `DISPATCHER` summary resolves calls by assuming that the receiver address
+is one of the contracts in the scene that implements the called method.  It
+will try every option, and if any of them can cause a counterexample, it will
```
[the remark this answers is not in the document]
mdgeorge4153: fixed

docs/user-guide/multicontract/index.md
```diff
@@ -0,0 +1,743 @@
...
+  choice; see {ref}`dispatcher` in the reference manual for full details.
+
+This summary means that when the Prover encounters an external call to
+`receiver.executeOperation(...)`, it will try to construct counterexamples
+where the `receiver` contract is any of the contracts in the scene that
+implement the `executeOperation` method.
+
+So far, there are no contracts in the scene that implement the
+`executeOperation` method, so the Prover will conservatively use a havoc
+summary for the call, and the rule will still fail.  To make use of the
```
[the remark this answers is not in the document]
mdgeorge4153: I'm pretty sure I ran a test and described what I saw.

docs/user-guide/multicontract/index.md
```diff
@@ -0,0 +1,743 @@
...
+[pool-link-script]: https://github.com/Certora/LiquidityPoolExample/blob/main/certora/scripts/verifyWithLinking.sh
+[flashloan-dispatcher]: https://github.com/Certora/LiquidityPoolExample/blob/main/certora/specs/flashLoan_dispatcher.spec
+[trivial-receiver]: https://github.com/Certora/LiquidityPoolExample/blob/main/certora/harness/TrivialReceiver.sol
+[trivial-script]:   https://github.com/Certora/LiquidityPoolExample/blob/main/certora/scripts/verifyFlashLoanTrivial.sh
+[transfer-receiver]: https://github.com/Certora/LiquidityPoolExample/blob/main/certora/harness/TransferReceiver.sol
+[transfer-script]:   https://github.com/Certora/LiquidityPoolExample/blob/main/certora/scripts/verifyFlashLoanTransfer.sh
+[arbitrary-values]:  https://github.com/Certora/LiquidityPoolExample/blob/main/certora/helpers/ArbitraryValues.sol
+[flexible-contract]: https://github.com/Certora/LiquidityPoolExample/blob/main/certora/harness/FlexibleReceiver.sol
+[flexible-linked]:   https://github.com/Certora/LiquidityPoolExample/blob/main/certora/scripts/verifyFlexibleLinked.sh
+[helpers]:           https://github.com/Certora/LiquidityPoolExample/blob/main/certora/helpers/
```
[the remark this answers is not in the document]
mdgeorge4153: I got the final spec to run using all of them, so I decided to leave them in.  If we have trouble with them when we flesh out the example more in the future, I will remove them

PR #34, opened 2022-07-27
added documentation for globalTimeout and showInteralMethods

mdgeorge4153:


docs/ref-manual/cli/options.md
```diff
@@ -479,4 +484,23 @@ This list is incomplete.
 This option determines whether {ref}`havoc summaries <havoc-summary>` assume
 that the called method returns the correct number of return values.
 
+(-showInternalFunctions)=
+#### `--settings -showInternalFunctions`
+
+This option causes the Prover to output a list of all the potentially
+summarizable internal functions.  In some cases the Prover is unable to locate
+all internal functions, and so summaries may not be applied.  This option can
+be useful to determine whether summary is applied or not.
```
jtoman, line 493: It's maybe worth nothing that output this is applied on a per-external method basis, i.e., it's plausible (but unlikely?) that within external contract function `foo(...)` we found an call to internal function `gorp(uint)` but in another external contract function `bar(...)` we did not. Further, we do have a rough call tree relationship, if `gorp(uint)` calls the internal function `frobulate(uint)`, this is also reflected in the output.

Again, I don't know how much detail you *want* to go into here: what you have is correct, but (imho) incomplete.

docs/ref-manual/cli/options.md
```diff
@@ -479,4 +484,23 @@ This list is incomplete.
...
+#### `--settings -globalTimeout=<seconds>`
+
+This option sets the global timeout in seconds.  By default, there is no global
+timeout.
+
+The global timeout is different from the {ref}`--smt_timeout` option: the
+`--smt_timeout` flag constrains the amount of time allocated to the processing
+of each individual rule, while the `-globalTimeout` flag constrains the
+processing of the entire job, including static analysis and other
+preprocessing.
```
jtoman, line 505: Worth nothing that this external timeout is enforced somewhat brutally: if we hit it, we try to cancel things reasonably but we will just hard stop any jobs that exceed the global timeout.

jtoman (commented):
Nothing *wrong* _per se_, just some TIOLI elaborations on the documented behavior

docs/ref-manual/cli/options.md
mdgeorge4153: I added these details.

docs/ref-manual/cli/options.md
mdgeorge4153: Here's what I wrote: "Jobs that exceed the global timeout will simply be terminated, so the result
reports may not be generated."  Does that address what's missing?

mdgeorge4153:
Note: the smt_timeout didn't seem to work the way I described it: https://vaas-stg.certora.com/output/93493/ca32709048bac27fee58/?anonymousKey=075c1369269e1242d58dd3dd63844d5707fa6bfd was run with --smt_timeout=300 but the rules are running for over 400s

docs/ref-manual/cli/options.md
nd-certora: I wasn't aware we have the globaltimeout feature. Is this the 2 hour hard stop?  If so there is a default...
Also are we sure we want to give it to users? I can imagine Silo giving an extremely large number. The whole reason we have the 2 hour hard stop is to save the computing resources

mdgeorge4153: I don't believe in security by obscurity.  I think if the feature's there, it should be documented.  If we don't want users using it, we should either remove the option, or use access control if we only want internal users to use it

docs/ref-manual/cli/options.md
```diff
@@ -479,4 +484,36 @@ This list is incomplete.
 This option determines whether {ref}`havoc summaries <havoc-summary>` assume
 that the called method returns the correct number of return values.
 
+(-showInternalFunctions)=
+#### `--settings -showInternalFunctions`
```
mdgeorge4153, line 488: Note: need to document where this output goes

docs/ref-manual/cli/options.md
mdgeorge4153: Note to self: I'm running a test to see if this works ([short timeout](https://vaas-stg.certora.com/output/6554/52e5c988e0d35207e737/?anonymousKey=508a15be19e7e55109608ef1fd489a2346c33917), [long timeout](https://vaas-stg.certora.com/output/6554/4814336bd9db1527859f/?anonymousKey=b1ab300f346fb056ab8595836d968fa356f554aa)).  I'll recheck on this when it's done.

mdgeorge4153: The short timeout did work, the long timeout didn't run long enough to hit the 2hr mark; I'm trying again with a longer `--smt_timeout`: https://vaas-stg.certora.com/output/6554/467f4209e300b03b850d?anonymousKey=092425fc311689830b985c3f1947767f0ea0ebf3

mdgeorge4153: Thanks to @0xDatapunk for running a test; the `--globalTimeout` did not work when set to longer than 2h.  Will update

docs/ref-manual/cli/options.md
mdgeorge4153: fixed

nd-certora (approved):
small fix

docs/ref-manual/cli/options.md
```diff
@@ -351,11 +351,21 @@ By default, we do not use a cache. If you want to use a cache to speed up the bu
...
+The Certora Prover generates a logical formula from the specification and
+source code. Then, it passes it on to an array of SMT solvers. The time it can
+take for the SMT solvers to solve the equation is highly variable, and could
+potentially be infinite. This is why they must be limited in run time.
+
+Note that the SMT timeout applies separately to each individual rule (or each method
+for parametric rules).  To set the global timeout, see {ref}`-globalTimeout`.
 
 **When to use it?**  
 The default time out for the solvers is 600 seconds. There are two use cases for this option.
```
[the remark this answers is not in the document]
mdgeorge4153: fixed

PR #38, opened 2022-08-18
Adding description of contract creation options

shellygr:


docs/ref-manual/cli/options.md
```diff
@@ -399,6 +399,53 @@ Where `TokenPair` is
 We have two contracts `BankToken.sol` and `LoanToken.sol`. We want `tokenA` of the `tokenPair` to be `BankToken`, and `tokenB` to be `LoanToken`. Addresses take up only one slot. We assume `tokenPair` is the first field of Bank (so it starts at slot zero). To do that, we use:  
 `certoraRun Bank.sol BankToken.sol LoanToken.sol --verify Bank:Bank.spec --structLink Bank:0=BankToken Bank:1=LoanToken`
 
+Options for controlling contract creation
+-----------------------------------------
+
+### `--dynamic_bound`
```
mdgeorge4153, line 405: I might make this `--dynamic_bound <n>` so that it's clear that it takes an argument

docs/ref-manual/cli/options.md
```diff
@@ -399,6 +399,53 @@ Where `TokenPair` is
 We have two contracts `BankToken.sol` and `LoanToken.sol`. We want `tokenA` of the `tokenPair` to be `BankToken`, and `tokenB` to be `LoanToken`. Addresses take up only one slot. We assume `tokenPair` is the first field of Bank (so it starts at slot zero). To do that, we use:  
 `certoraRun Bank.sol BankToken.sol LoanToken.sol --verify Bank:Bank.spec --structLink Bank:0=BankToken Bank:1=LoanToken`
 
+Options for controlling contract creation
```
mdgeorge4153, line 402: running `make` returned 4 spelling errors; can you either correct them or add them to the dictionary?

docs/ref-manual/cli/options.md
```diff
@@ -399,6 +399,53 @@ Where `TokenPair` is
...
+-----------------------------------------
+
+### `--dynamic_bound`
+
+**What does it do?**
+If set to zero (the default) dynamic creation will not be modeled. If non-zero, then dynamic contract creation will be modeled with cloning, where each contract will be cloned at most n times.
+**When to use it?**
```
mdgeorge4153, line 409: The convention elsewhere in the document is to put a paragraph break between these sections

docs/ref-manual/cli/options.md
```diff
@@ -399,6 +399,53 @@ Where `TokenPair` is
...
+When you wish to model contract creation as a non-havoc.
+**Example**
+Suppose a contract `C` creates a new instance of a contract `Foo`, and you wish to inline the constructor of `Foo` at the creation site.
+`certoraRun C.sol Foo.sol --dynamic_bound 1`
+
+### `--dynamic_dispatch`
+
+**What does it do?**
+If false (the default), then all contract method invocations on newly created instances on spec must be modeled with explicit DISPATCHER summaries.
```
mdgeorge4153, line 418: code font for `DISPATCHER`

docs/ref-manual/cli/options.md
```diff
@@ -399,6 +399,53 @@ Where `TokenPair` is
 We have two contracts `BankToken.sol` and `LoanToken.sol`. We want `tokenA` of the `tokenPair` to be `BankToken`, and `tokenB` to be `LoanToken`. Addresses take up only one slot. We assume `tokenPair` is the first field of Bank (so it starts at slot zero). To do that, we use:  
 `certoraRun Bank.sol BankToken.sol LoanToken.sol --verify Bank:Bank.spec --structLink Bank:0=BankToken Bank:1=LoanToken`
 
+Options for controlling contract creation
+-----------------------------------------
+
+### `--dynamic_bound`
+
+**What does it do?**
+If set to zero (the default) dynamic creation will not be modeled. If non-zero, then dynamic contract creation will be modeled with cloning, where each contract will be cloned at most n times.
```
mdgeorge4153, line 408: What does it mean to not be modeled?  Does it mean the `new` statement is rejected, treated as a havoc, ignored, ...?

mdgeorge4153, line 408: It's probably worth mentioning `new` somewhere here, it's not clear that a user will know what "dynamic contract creation" is.

docs/ref-manual/cli/options.md
mdgeorge4153, line 418: What does "on spec" mean here?

docs/ref-manual/cli/options.md
```diff
@@ -399,6 +399,53 @@ Where `TokenPair` is
...
+When you wish to model contract creation as a non-havoc.
+**Example**
+Suppose a contract `C` creates a new instance of a contract `Foo`, and you wish to inline the constructor of `Foo` at the creation site.
+`certoraRun C.sol Foo.sol --dynamic_bound 1`
+
+### `--dynamic_dispatch`
+
+**What does it do?**
+If false (the default), then all contract method invocations on newly created instances on spec must be modeled with explicit DISPATCHER summaries. 
+If true, then Prover will, on a best effort basis, automatically apply the DISPATCHER summary for call-sites that must be with a newly created contract as a receiver.
```
mdgeorge4153, line 419: **the** Prover, and "best-effort"

mdgeorge4153: "call sites"

docs/ref-manual/cli/options.md
mdgeorge4153, line 418: what do we mean by "must be modeled"?  I think maybe it should be "must be summarized".  Actually, we should make it clearer: "the user must explicitly write `DISPATCHER` summaries for all methods called on newly created instances".

docs/ref-manual/cli/options.md
```diff
@@ -399,6 +399,53 @@ Where `TokenPair` is
...
+**When to use it?**
+When you prefer not to add explicit `DISPATCHER` summaries to methods invoked by the created contract.
+**Example**
+Suppose a contract `C` creates a new instance of a contract `Foo`, and you wish to inline the constructor of `Foo` at the creation site, 
+and `Foo` calls some method `m()` which you wish to automatically link to the newly created contract.
+`certoraRun C.sol Foo.sol --dynamic_bound 1 --dynamic_dispatch true`
```
mdgeorge4153, line 425: Does `--dynamic_dispatch` require `--dynamic_bound`?  If so, we should mention it.

docs/ref-manual/cli/options.md
mdgeorge4153, line 418: I think the first mention of `DISPATCHER` summaries should be a link to the corresponding page.  I think you can write ``` {ref}`` `DISPATCHER` <dispatcher>`` ``` To make a reference with code font text.

docs/ref-manual/cli/options.md
```diff
@@ -399,6 +399,53 @@ Where `TokenPair` is
...
+
+### `--prototype`
```
mdgeorge4153, line 427: `--prototype <hex string>=<contract>`

docs/ref-manual/cli/options.md
```diff
@@ -399,6 +399,53 @@ Where `TokenPair` is
...
+
+**What does it do?**
+Instructs the Prover to consider an hexadecimal string passed into a `create` or `create2` call as returning an instance of a specific contract.
```
mdgeorge4153, line 430: The string doesn't return anything.  I think "Instructs the Prover to use a specific contract type for the return value from a call to `create` or `create2` on the given hex string" is correct, but I'm not sure.

docs/ref-manual/cli/options.md
```diff
@@ -399,6 +399,53 @@ Where `TokenPair` is
...
+**When to use it?**
+If you have contract creation that uses low level calls to `create` or `create2`.
```
mdgeorge4153, line 432: If you are verifying a contract that uses low level calls to `create` or `create2` for contract creation.

docs/ref-manual/cli/options.md
```diff
@@ -399,6 +399,53 @@ Where `TokenPair` is
...
+**Example**
+Suppose you have a contract `C` that creates another contract `Foo` like this:
+```
```
mdgeorge4153, line 435: use
````
```solidity
...
````

File: docs/ref-manual/cli/options.md [hunk 12]
```diff
@@ -399,6 +399,53 @@ Where `TokenPair` is
...
+     mstore(ptr, 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000000000000000000000)
+     mstore(add(ptr, 0x14), shl(0x60, implementation))
+     mstore(add(ptr, 0x28), 0x5af43d82803e903d91602b57fd5bf30000000000000000000000000000000000)
+     instance := create(0, ptr, 0x37)
+}
+```
+Then you can set the string `3d602d80600a3d3981f3363d3d373d3d3d363d73` appearing in the first `mstore` after the `0x` prefix as a "prototype" for `Foo`. 
+The Prover will then be able to create a new instance of `Foo` at the point where the code creates it:
+`certoraRun C.sol Foo.sol --prototype 3d602d80600a3d3981f3363d3d373d3d3d363d73=Foo --dynamic_bound 1`
+Note: this argument has no effect if the dyanmic_bound is zero.
```
mdgeorge4153, line 447: `dynamic_bound` is misspelled and not in code font.

mdgeorge4153: In fact, I might just replace this with a link: add `(--dynamic_bound)=` on the line before the section header for the `--dynamic_bound`.  Then you can use ``{ref}`--dynamic_bound` `` to make a link.  You can also use ``{ref}`dynamic bound <--dynamic bound>` `` to use "dynamic bound" as the link text.

docs/ref-manual/cli/options.md
mdgeorge4153: (actually, it looks like 3 of them were introduced in an earlier PR that I forgot to run spell checking on)

mdgeorge4153: (the dictionary is `spelling_wordlist.txt` in the root directory)

mdgeorge4153 (changes requested):
Thanks!  I left a bunch of editorial comments that would be good to fix up before merging

docs/ref-manual/cli/options.md
jtoman: say "contract creation (via the `new` statement or the `create/create2` instructions)" or something

docs/ref-manual/cli/options.md
```diff
@@ -399,6 +399,53 @@ Where `TokenPair` is
...
 
+Options for controlling contract creation
+-----------------------------------------
+
+### `--dynamic_bound`
+
+**What does it do?**
+If set to zero (the default) dynamic creation will not be modeled. If non-zero, then dynamic contract creation will be modeled with cloning, where each contract will be cloned at most n times.
+**When to use it?**
+When you wish to model contract creation as a non-havoc.
```
jtoman, line 410: I think that modelling does more than `non-havoc`, it simulates the actual creation of the contract as a full-fledged entity on our (fake) block chain. It may be worth mentioning the behavior *without* this turned on: "without it, the `create` and `create2` commands simply return a fresh address; the prover does not model their storage, code, constructors, immutables, etc. Any interaction with these generated addresses is modeled imprecisely with conservative havoc."

docs/ref-manual/cli/options.md
jtoman: It's worth noting the limitations of this flag: it *only* does something when the prover can *prove* that the callee is a dynamically created contract. That is, if we have:

```
MyFoo f;
if(*) {
   f = new MyFoo(...);
} else {
  f = storageStruct.myFoo;
}
f.bar();
```

the automatic dispatcher won't do anything

docs/ref-manual/cli/options.md
jtoman: yes

jtoman: worth showing a code example where this *doesn't* work imo

docs/ref-manual/cli/options.md
```diff
@@ -399,6 +399,53 @@ Where `TokenPair` is
...
+assembly {
+     let ptr := mload(0x40)
+     mstore(ptr, 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000000000000000000000)
+     mstore(add(ptr, 0x14), shl(0x60, implementation))
+     mstore(add(ptr, 0x28), 0x5af43d82803e903d91602b57fd5bf30000000000000000000000000000000000)
+     instance := create(0, ptr, 0x37)
+}
+```
+Then you can set the string `3d602d80600a3d3981f3363d3d373d3d3d363d73` appearing in the first `mstore` after the `0x` prefix as a "prototype" for `Foo`. 
+The Prover will then be able to create a new instance of `Foo` at the point where the code creates it:
```
jtoman, line 445: worth mentioning that the hex string must be:
* a strict prefix of the memory region passed to the `create` command
* must be unique within each invocation of the tool
* must not contain gaps, i.e., `3d602d80600a3d3981f3363d3d373d3d3d363d730000` will not work (those last four bytes will be overwritten) but `3d602d80600a3d3981f3363d3d373d3d3d363d` will

Also extremely important to note that the constructor code for `Foo` is *not* called in this case.

docs/ref-manual/cli/options.md
shellygr: not true but I'll apply everywhere

docs/ref-manual/cli/options.md
shellygr: rephrased

docs/ref-manual/cli/options.md
shellygr: not sure what you mean here John

jtoman: Showing an example where the best effort heuristic fails.

docs/ref-manual/cli/options.md
```diff
@@ -444,7 +485,13 @@ assembly {
 Then you can set the string `3d602d80600a3d3981f3363d3d373d3d3d363d73` appearing in the first `mstore` after the `0x` prefix as a "prototype" for `Foo`. 
 The Prover will then be able to create a new instance of `Foo` at the point where the code creates it:
 `certoraRun C.sol Foo.sol --prototype 3d602d80600a3d3981f3363d3d373d3d3d363d73=Foo --dynamic_bound 1`
-Note: this argument has no effect if the dyanmic_bound is zero.
+Note: this argument has no effect if the {ref}`dynamic bound <--dynamic_bound>` is zero.
+
+Also note that the hex string must be: 
+- a strict prefix of the memory region passed to the create command
+- must be unique within each invocation of the tool
+- must not contain gaps, i.e., `3d602d80600a3d3981f3363d3d373d3d3d363d730000` will not work (those last four bytes will be overwritten) but `3d602d80600a3d3981f3363d3d373d3d3d363d` will
```
jtoman, line 493: "e.g., `....730000` in the  above example will not work"

jtoman (approved):
lgtm, leaving final approval for Mike

mdgeorge4153 (approved):
looks good, will merge

PR #39, opened 2022-09-08
Mike/induction

mdgeorge4153:
Several people have asked when/whether it is safe to use `requireInvariant` inside of a `preserved` block.  I checked it for myself and wrote up a proof.

Link to generated documentation: https://certora-certora-prover-documentation--39.com.readthedocs.build/en/39/docs/ref-manual/cvl/invariants.html#invariants-and-induction


docs/ref-manual/cvl/statements.md
```diff
@@ -70,6 +70,10 @@ This section is incomplete.  See [the old documentation](/docs/confluence/anatom
 This feature is currently undocumented.
 ```
 
+```{todo}
+`requireInvariant` is always safe, even in `preserved` blocks; see {ref}`invariant-induction`.
```
nd-certora, line 74: for an invariant that is fully proved

nd-certora (approved):

gadi03:
Regarding "apply any constructor to the uninitialized contract" - when checking constructors the prover checks only initialized contracts, that is, all state variables are initialized to zeros, or whatever their default values are. EVM promises that variables are initialized with their default values, so those are the concrete values. So we can concludes that the proof of the constructors is still valid.  
It's a bit confusing - when checking an invariant on a method the prover assigns variables with nondeterministic values.

gadi03 (commented):
Regarding "apply any constructor to the uninitialized contract" - when checking constructors the prover checks only initialized contracts, that is, all state variables are initialized to zeros, or whatever their default values are. EVM promises that variables are initialized with their default values, so those are the concrete values. So we can concludes that the proof of the constructors is still valid.
It's a bit confusing - when checking an invariant on a method the prover assigns variables with nondeterministic values.

gadi03:
Hi, please review the example below. Invariant `i` and invariant `j` fail, but when `requireInvariant j` is added to `i` and  `requireInvariant i` is added to `j` they pass.

https://github.com/Certora/CustomersCode/blob/invariant_example/invariant_example/test_spec.spec

https://prover.certora.com/output/99352/de90d2d48b0cfa96a8ff?anonymousKey=f833a599ff613f172c4e8d10566aaf38e7654a09

gadi03 (approved):

mdgeorge4153:
> 

Note: all arithmetic in this comment is done mod 6.

Invariant i evaluates to true on 4 and 5 (mod 6), and false otherwise.  `double_plus_1` takes the contract from state 4 to state 3, so this takes a good state to a bad state, and the invariant should fail.

Similarly, invariant j would evaluate to true if x is 0, 1, or 5, but would evaluate to false if x is 2, 3, or 4 (mod 6).  `double_plus_1` in state 1 takes to contract to state 3, so again the contract can move from a good state to a bad state, and the invariant should fail.

However, `i && j` is true only in state 5, and `double_plus_1` takes 5 to 5 (and the contract starts in state 5), so the rule should pass.

Notably, the contract is always in state 5, so both invariants do actually hold on both reachable states.

Do you think there needs to be something added to this document to explain this situation?  I expect to have an example like this in the user guide, but that hasn't been written yet.

docs/ref-manual/cvl/statements.md
mdgeorge4153: fixed

PR #40, opened 2022-11-02
fix --path option

rahav-certora:
The --path option is passed as is to the solc compiler as the --allow-paths option.

### DIFF
--- a/docs/ref-manual/cli/options.md
+++ b/docs/ref-manual/cli/options.md
@@ -264,8 +264,8 @@ When different contracts have to be compiled for different Solidity versions.
 ### `--path`
 
 **What does it do?**  
-Passes the value of this option as it to the solc option `--allow-paths`
-see [--allow-path specification](https://docs.soliditylang.org/en/v0.8.16/path-resolution.html#allowed-paths)
+Passes the value of this option as is to the solidity compiler's option `--allow-paths`.
+See [--allow-path specification](https://docs.soliditylang.org/en/v0.8.16/path-resolution.html#allowed-paths)
 
 **When to use it?**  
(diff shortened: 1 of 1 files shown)


docs/ref-manual/cli/options.md
```diff
@@ -264,11 +264,13 @@ When different contracts have to be compiled for different Solidity versions.
 ### `--path`
 
 **What does it do?**  
-Use the given path as the root of the source tree instead of the root of the file system.  
+Passes the value of this option as it to the solc option `--allow-paths`
```
mdgeorge4153, line 267: `solc` should be code font.

mdgeorge4153, line 267: "as it" => "as it is"

docs/ref-manual/cli/options.md
```diff
@@ -264,11 +264,13 @@ When different contracts have to be compiled for different Solidity versions.
...
+see [--allow-path specification](https://docs.soliditylang.org/en/v0.8.16/path-resolution.html#allowed-paths)
```
mdgeorge4153, line 268: capitalize, period.

mdgeorge4153 (approved):
I added some small nitpicks.  Feel free to merge after you address them

docs/ref-manual/cli/options.md
rahav-certora: solc is not the right word, changed it to "solidity compiler"

rahav-certora: if fact I wanted to write "as is", does that work?

docs/ref-manual/cli/options.md
rahav-certora: done

PR #47, opened 2022-12-18
use correct 'foo.selector' syntax in examples.

Melvillian:
Prior to this commit, there were 2 instances in the documentation using the old syntax for getting a method's select (e.g. 'foo.selector()'). Now those are using the correct syntax, which does not use the function-like open + closed parentheses.


mdgeorge4153 (approved):
Looks  good, thanks!

PR #51, opened 2022-12-28
fix example in old documentation's Definitions section.

Melvillian:
prior to this commit, the type for the unquantified variable in the 'exists' example was not defined, which is not valid .spec syntax.


mdgeorge4153 (approved):
looks good, thank you

PR #52, opened 2023-02-01
DOC-305 Update sanity.md

mailalexjoseph:
documentation for advanced sanity check

Generated docs: https://certora-certora-prover-documentation--52.com.readthedocs.build/en/52/docs/prover/checking/sanity.html


docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,243 @@ Rule Sanity Checks
...
+     assert f(x) == 2, "f must return 2";
+   }
+   ```
+   Since there are no models satisfying both `x > 2` and `x < 1`, this rule
+   will always pass, regardless of the behavior of the contract.  This is an
+   example of a *vacuous* rule - one that passes only because the preconditions
+   are contradictory.
+
+   ```{caution}
+   The reachability check will *pass* on vacuous rules and *fail* on correct
```
nd-certora, line 30: this is no longer true, right? it used to be that we showed that the failing rule on correct rules

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,243 @@ Rule Sanity Checks
...
+   ```
+  
+  
+   Since every `uint` satisfies the assertion, the assertion is tautological,
+   which is likely to be an error in the specification.
+   
+   **Checking vacuity for invariants**
+   
+   For invariant, vacuity is checked by converting it into a rule without any requires.
+
```
nd-certora, line 61: need more explanations here

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,243 @@ Rule Sanity Checks
...
+   
+   
+    ```cvl
+    
+      // Not a tautology
+      invariant rootNonZero()
+         root() != 0
+
+      rule rootNonZeroRule(){
+         assert root() != 0;
```
nd-certora, line 85: show a run? more explanation. since it fails it is not a tautology, but this is not clear

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,243 @@ Rule Sanity Checks
...
+    
+    For rules, checking for tautology requires checking each assertion to see if 
+    it’s meaningful. In order to do this, we employ few different checks depending
+    on the syntax of the assertion expression.
+    
+      *Case 1: Implication*
+    
+      Given a rule with an `assert p => q` we perform two checks:
+    
+      1. `assert(!p)`
```
nd-certora, line 108: let's give names to each check, so then in the ui (next release), we will give links to the doc from the error message
Implication Hypothesis

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,243 @@ Rule Sanity Checks
...
+    2. `assert(q)`
+       
+       If the conclusion part is always true regardless of the hypothesis then the
+       assertion is a tautology
+       
+        ```cvl
+          rule testSanity{
+          uint a;
+          uint b;
+          assert a>10 => b>0;
```
nd-certora, line 140: b>=0

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,243 @@ Rule Sanity Checks
...
+        
+      ```cvl
+            assert-vacuity check FAILED: sanity.spec:11:5
+            assert-tautology check FAILED: sanity.spec:11:5'a < 0 => b < 10' is a vacuous 
+            implication. It could be rewritten to !a < 0 because a < 0 is always false
+            
+      ```
+      
+        
+    2. `assert(q)`
```
nd-certora, line 131: Implication-Conclusion

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,243 @@ Rule Sanity Checks
...
+              }
+            ```
+            
+      Error message
+            
+      ```cvl
+            assert-tautology check FAILED: sanity.spec:47:5the expression `b >= 0` is always true
+      ```
+            
+   **Other advance checks for rules currently unavailable**
```
nd-certora, line 231: remove

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+      Error message
+           
+      ```cvl
+           assert-tautology check FAILED: sanity.spec:26:5'a < 0 <=> b < 0' could be rewritten 
+           to !a < 0 && !b < 0 because both a < 0 and b < 0 are always false
+      ```
+           
+      2. Double implication, both true: `assert(p && q)`
+      
+          If this passes then the is a tautology since both conditions are always true.
```
nd-certora, line 177: the rule

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
 ==================
 
 The {ref}`--rule_sanity` option enables some automatic checks that can warn you
```
Dmwihuri, line 4: The {ref}`--rule_sanity` option enables some automatic checks that warn users

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+may be followed by one of `none`, `basic`, or `advanced` options.
+
+There are 3 kinds of sanity checks:
+
+1. **Reachability** checks that even when ignoring all the user-provided
+   assertions, the end of the rule is reachable. This check ensures that that
+   the combination of `require` statements does not rule out all possible
+   counterexamples.
+
+   For example, the following rule would be flagged by the reachability check:
```
Dmwihuri, line 15: _Example_
The following rule would be flagged by the reachability check:

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+   rule vacuous {
+     uint x;
+     require x > 2;
+     require x < 1;
+     assert f(x) == 2, "f must return 2";
+   }
+   ```
+   Since there are no models satisfying both `x > 2` and `x < 1`, this rule
+   will always pass, regardless of the behavior of the contract.  This is an
+   example of a *vacuous* rule - one that passes only because the preconditions
```
Dmwihuri, line 26: example of a *vacuous* rule  (one that passes only because the preconditions
   are contradictory).

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+   are contradictory.
+
+   ```{caution}
+   The reachability check will *pass* on vacuous rules and *fail* on correct
```
Dmwihuri, line 30: The reachability check *passes* on vacuous rules and *fail* on correct

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+   rules.  A passing reachability check indicates a potential error in the rule.
+   
+   The exception is when a {term}`parametric rule` is checked on the default
+   fallback function: The default fallback function should always revert, so
+   there are no examples that can reach the end of the rule.
```
Dmwihuri, line 35: that no examples can reach the end of the rule.

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+   ```
+   
+2. **Assert-Vacuity** checks that individual `assert` statements are not
+   tautologies.  A tautology is a statement that is true on all examples, even
```
Dmwihuri, line 39: A tautology is defined as a statement that is true on all examples, even
   if all the `require` and `if` conditions are removed.

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+   if all the `require` and `if` conditions are removed.
+
+   For example, the following rule would be flagged by the assert-vacuity check:
```
Dmwihuri, line 42: _Example_
The following rule would be flagged by the assert-vacuity check:

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+     uint x; uint y;
+     require x != y;
+     ...
+     assert x < 2 || x >= 2,
+      "x must be smaller than 2 or greater than or equal to 2";
+   }
+   ```
+  
+  
+   Since every `uint` satisfies the assertion, the assertion is tautological,
```
Dmwihuri, line 55: Since every `unit` satisfies the assertion, the assertion is tautological. This means that there is a likelihood of an error in the specification.

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
...tate variables `address root` and `uint a`. Due to the zero address checks in the `constructor` and the `changeRoot` function, the root address can never be zero. The `rootNonZero` invariant asserts that the root address is never zero. When we run this invariant with the `--rule_sanity` `advanced` or `basic` options, the prover creates a rule similar to the `rootNonZeroRule` below. This rule would fail since the tool could assume a starting state where the `root` is 0. This means that the invariant expression is not a tautology and the invariant passes. On the other hand the `aGE0` invariant, when run without the `--rule_sanity` option, will pass [view report](https://vaas-stg.certora.com/output/11775/871cf37193c75d27542b/?anonymousKey=dde443c4a806021716e863a454561a6ad1543d2e) but when we run it with the `--rule_sanity` `advanced` or `basic` options, the prover creates a rule similar to the `aGE0Rule` below. This rule passes, indicating that the invariant expression is a tautology. The verification report shows that the invariant failed vacuity check [view report](https://vaas-stg.certora.com/output/11775/4c4cb65f65c75f013c63/?anonymousKey=0b6a843857e6ead8e1bb1f11b984fb6e3e9fb6a8).
```
Dmwihuri, line 62: The `sanityCheck` contract below has two state variables:
1. `address root` and 
2. `uint a`
Due to the zero address checks in the `constructor` and the `changeRoot` function, the root address can never be zero. The `rootNonZero` invariant asserts that the root address is never zero. When this invariant is run  with the `--rule_sanity` `advanced` or `basic` options, the prover creates a rule similar to the `rootNonZeroRule` below. This rule would fail since the tool could assume a starting state where the `root` is 0. This means that the invariant expression is not a tautology and the invariant passes. On the other hand, the `aGE0` invariant, when run without the `--rule_sanity` option, will pass [view report](https://vaas-stg.certora.com/output/11775/871cf37193c75d27542b/?anonymousKey=dde443c4a806021716e863a454561a6ad1543d2e), but when we run it with the `--rule_sanity` `advanced` or `basic` options, the prover creates a rule similar to the `aGE0Rule` below. This rule passes, indicating that the invariant expression is a tautology. The verification report below shows that the invariant failed vacuity check
 [view report](https://vaas-stg.certora.com/output/11775/4c4cb65f65c75f013c63/?anonymousKey=0b6a843857e6ead8e1bb1f11b984fb6e3e9fb6a8).

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+
+      rule aGE0Rule{
+         assert a() >= 0;
+      }
+
+    ```
+    
+    **Checking vacuity for rules**
+    
+    For rules, checking for tautology requires checking each assertion to see if
```
Dmwihuri, line 102: When working with rules, checking for tautology requires inspecting each assertion to see if 
    it’s meaningful. To do this, we employ few different checks depending
    on the syntax of the assertion expression.

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+      Error Message
+        
+      ```cvl
+              assert-tautology check FAILED: sanity.spec:21:5conclusion `b >= 0` is always true 
+              regardless of the hypothesis
+      ```
+        
+   *Case 2: Double Implication*
+     
+     Given a rule with an assert p <=> q we perform two checks:
```
Dmwihuri, line 155: Given a rule with an assert p <=> q, we perform two checks:

Dmwihuri (changes requested):
Hi, I'm suggesting a fee changes here. Kindly review

mdgeorge4153:
We should also update the `sanity` section of `glossary.md`, which currently has a placeholder

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+    it’s meaningful. In order to do this, we employ few different checks depending
+    on the syntax of the assertion expression.
+    
+      *Case 1: Implication*
+    
+      Given a rule with an `assert p => q` we perform two checks:
+    
+      1. Implication hypothesis: `assert(!p)`
+       
+      If the hypothesis part is always false then the assertion is a tautology.
```
CowardlyBattleCat, line 112:
```suggestion
      If the hypothesis part is always false then the implication can never evaluate to false and the assertion is a tautology.
```

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+      ```cvl
+              rule testSanity{
+              uint a;
+              uint b;
+              assert a<0 => b<10;
+              }
+      ```
+         
+        
+      Error message
```
CowardlyBattleCat, line 123:
```suggestion
      Failing a check of this kind will yield an error message like the following:
```

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+       
+        ```cvl
+          rule testSanity{
+          uint a;
+          uint b;
+          assert a>10 => b>=0;
+          }
+        ```
+        
+      Error Message
```
CowardlyBattleCat, line 146:
```suggestion
      Failing a check of this kind will yield an error message like the following:
```

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+            assert-tautology check FAILED: sanity.spec:11:5'a < 0 => b < 10' is a vacuous 
+            implication. It could be rewritten to !a < 0 because a < 0 is always false
+            
+      ```
+      
+        
+    2. Implication conclusion: `assert(q)`
+       
+       If the conclusion part is always true regardless of the hypothesis then the
+       assertion is a tautology
```
CowardlyBattleCat, line 136:
```suggestion
       If the conclusion is always true regardless of the hypothesis then the
       implication can never evaluate to false and the assertion is a tautology.
```

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+
+           ```cvl
+             rule sanityDoubleImplication1{
+             uint a;
+             uint b;
+             assert a<0 <=> b<0;
+             }
+           ```
+          
+      Error message
```
CowardlyBattleCat, line 168:
```suggestion
      Failing a check of this kind will yield an error message like the following:
```

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+      
+            ```cvl
+              rule sanityDoubleImplication2{
+              uint a;
+              uint b;
+              assert a>=0 <=> b>=0;
+              }
+            ```
+           
+     Error message
```
CowardlyBattleCat, line 187:
```suggestion
     Failing a check of this kind will yield an error message like the following:
```

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+
+            ```cvl
+              rule sanityDisjunction1{
+              uint a;
+              uint b;
+              assert a>=0 || b>10;
+              }
+            ```
+           
+      Error message
```
CowardlyBattleCat, line 209:
```suggestion
      Failing a check of this kind will yield an error message like the following:
```

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+          
+            ```cvl
+              rule sanityDisjunction2{
+              uint a;
+              uint b;
+              assert a>10 || b>=0;
+              }
+            ```
+            
+      Error message
```
CowardlyBattleCat, line 227:
```suggestion
      Failing a check of this kind will yield an error message like the following:
```

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+              assert-tautology check FAILED: sanity.spec:21:5conclusion `b >= 0` is always true 
+              regardless of the hypothesis
+      ```
+        
+   *Case 2: Double Implication*
+     
+     Given a rule with an assert p <=> q we perform two checks:
+     
+     1. Double implication, both false: `assert(!p && !q)`
+         If this passes then the assertion is a tautology since both conditions are always false.
```
CowardlyBattleCat, line 158:
```suggestion
         If this passes then the assertion is a tautology since both conditions are always false leaving no way for the biconditional statement to evaluate to false.
```

docs/prover/checking/sanity.md
CowardlyBattleCat, line 177:
```suggestion
          If this passes then the rule is a tautology since both conditions are always true and the biconditional statement can never be false.
```

docs/prover/checking/sanity.md
CowardlyBattleCat, line 155:
```suggestion
     Given a rule with an `assert p <=> q` we perform two checks:
```

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+     Error message
+            
+     ```cvl
+            assert-tautology check FAILED: sanity.spec:33:5'a >= 0 <=> b >= 0' could be rewritten
+            to a >= 0 && b >= 0 because both a >= 0 and b >= 0 are always true
+     ```
+            
+   *Case 3: Disjunction*
+   
+      Given a rule with an assert p || q we perform two checks:
```
CowardlyBattleCat, line 196:
```suggestion
      Given a rule with an `assert p || q` we perform two checks:
```

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+      
+      1. Disjunction always true: `assert(p)`
+          If this passes then the assertion is a tautology since the first expression is always true.
```
CowardlyBattleCat, line 199:
```suggestion
          If this passes then the assertion is a tautology since the first expression of the disjunction is always true.
```

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+           
+      Error message
+            
+      ```cvl
+            assert-tautology check FAILED: sanity.spec:41:5the expression `a >= 0` is always true
+      ```
+            
+      2. Disjunction always true: `assert(q)`
+          
+          If this passes then the assertion is a tautology since the second expression is always true.
```
CowardlyBattleCat, line 217:
```suggestion
          If this passes then the assertion is a tautology since the second expression of the disjunction is always true.
```

docs/prover/checking/sanity.md
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+              }
+            ```
+            
+      Error message
+            
+      ```cvl
+            assert-tautology check FAILED: sanity.spec:47:5the expression `b >= 0` is always true
+      ```
+
+
```
CowardlyBattleCat, line 233:
````suggestion
3. **Require-Redundancy** checks for redundant `require` statements.
   A `require` is considered to be redundant if it can be removed without
   affecting the satisfiability of the rule.

   For example, the require-redundancy check would flag the following rule:
   ```cvl
   rule require_redundant {
     uint x;
     require x > 3;
     require x > 2;
     assert f(x) == 2, "f must return 2";
   }
   ```
   In this example, the second requirement is redundant, since any `x` greater
   than 3 will also be greater than 2.
````

File: docs/prover/checking/sanity.md [hunk 33]
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+            
+      Error message
+            
+      ```cvl
+            assert-tautology check FAILED: sanity.spec:47:5the expression `b >= 0` is always true
+      ```
+
+      
+       
+
```
CowardlyBattleCat, line 235:
````suggestion
```{todo}
Add details of specific Require-Redundancy checks if helpful
```
````

**Review by CowardlyBattleCat — CHANGES_REQUESTED** [review]
This is looking really good. 

I tried to reorder and rework a few places for clarity, but I didn't focus overly on careful formatting.

I also brought over the Require-Redundancy section from the CLI options page, meaning nearly all of the `--rule_sanity` details there are now present (and expanded) here.

File: docs/prover/checking/sanity.md [hunk 34]
```diff
@@ -2,9 +2,241 @@ Rule Sanity Checks
...
+      // Tautology
+      invariant aGE0()
+         a() >= 0
+
+      rule aGE0Rule{
+         assert a() >= 0;
+      }
+
+    ```
```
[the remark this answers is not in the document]
CowardlyBattleCat: For the paragraph dealing with the `aGE0` invariant:
* will it cause any issues later if we are linking directly to rule reports here?
* if we do link directly, should we link to production rather than staging?

docs/prover/checking/sanity.md
mdgeorge4153:
```suggestion
     Given a rule with an `assert p <=> q`, we perform two checks:
```

docs/prover/checking/sanity.md
mdgeorge4153:
```suggestion
When working with rules, checking for tautology requires inspecting each assertion to see if
it’s meaningful. To do this, we employ a few different checks depending
on the syntax of the assertion expression.
```

docs/prover/checking/sanity.md
```diff
@@ -1,10 +1,304 @@
...
+```
+    
+(sanity-assert-tautology)=
+Assert tautology checks
+-----------------------
+
+### Checking vacuity for rules
+    
+For rules, checking for tautology requires checking each assertion to see if 
+it’s meaningful. In order to do this, we employ few different checks depending
```
CowardlyBattleCat, line 154:
```suggestion
it’s meaningful. In order to do this, we employ a few different checks depending
```

docs/prover/checking/sanity.md
```diff
@@ -1,10 +1,304 @@
...
+   Error Message
+    
+   ```
+   assert-tautology check FAILED: sanity.spec:21:5conclusion `b >= 0` is always true 
+   regardless of the hypothesis
+   ```
+        
+#### Tautology checks for double implication
+
+Given a rule with an assert p <=> q we perform two checks:
```
CowardlyBattleCat, line 201:
```suggestion
Given a rule with an `assert p <=> q` we perform two checks:
```

CowardlyBattleCat (commented):
Added/changed backticks to be consistent with earlier usage (line 158)

docs/prover/checking/sanity.md
```diff
@@ -1,10 +1,314 @@
...
+   Error Message
+    
+   ```
+   assert-tautology check FAILED: sanity.spec:21:5conclusion `b >= 0` is always true 
+   regardless of the hypothesis
+   ```
+        
+#### Tautology checks for double implication
+
+Given a rule with an `assert p <=> q` we perform two checks:
```
CowardlyBattleCat, line 208:
```suggestion
Given a rule with an `assert p <=> q`, we perform two checks:
```

mdgeorge4153:
This is superseded by #71

PR #53, opened 2023-02-01
Document hashing in the Certora Prover Refererence Manual [DOC-166]

alexandernutz:
Includes
 - general information on how we model (bounded) hashing
 - documenting the new `--optimistic_hashing` and `--hashing_length_bound` flags
 - explanation on how we model unbounded hashing (by reduction to bounded hashing)

linked to https://certora.atlassian.net/browse/DOC-166

Generated documentation: https://certora-certora-prover-documentation--53.com.readthedocs.build/en/53/docs/prover/approx/hashing.html


docs/prover/approx/hashing.md
```diff
@@ -0,0 +1,182 @@
+Modeling of Hashing in CVT
+===========
+
+
+## Introduction
+
+The Keccak hash function is used heavily by Solidity smart contracts. 
+Most prominently, all unbounded data structures in storage (arrays, mappings) receive their storage addresses as values of the Keccak function.
```
mdgeorge4153, line 8: Maybe worth mentioning here or somewhere that this approximation is also used when users call the solidity keccak functions (assuming that that's correct).

conf.py
```diff
@@ -35,6 +35,7 @@
     'myst_parser',
     'sphinx_rtd_theme',
     'sphinx.ext.todo',
+    'sphinxcontrib.spelling',
```
mdgeorge4153, line 38: curious why you added this...was there a problem with the spell checking config?

docs/prover/approx/hashing.md
```diff
@@ -0,0 +1,182 @@
...
+
+The Certora Prover does not operate with an actual implementation of the Keccak hash function, since this would make most verification intractable and provide no benefits in all practical cases we are aware of so far.
```
mdgeorge4153, line 10: "...provide no practical benefits."

docs/prover/approx/hashing.md
```diff
@@ -0,0 +1,182 @@
...
+
+We can see that non-collision of hashes is essential for storage integrity. E.g., if `hash(1) + j` was equal to `hash(2, k)` then the operations on `a` an `m` would interfere with each other.
+
+Also note that the initial storage slots are reserved, i.e., we make sure that no hash value ends up colliding with slots 0 to 10000.
+
+
+## Modeling the Keccak Function (bounded case)
+
+The Certora Prover models the Keccak hash function as an arbitrary function that is _injective with large gaps_.
+That means that if `x != y` then `hash(x) != hash(y)`, but also that for all additive offsets `o` that actually occur in the program `hash(x) + o != hash(y)`.
```
mdgeorge4153, line 52: The `o` looks like a `0`.  Maybe `i` or `a` or `c`

docs/prover/approx/hashing.md
```diff
@@ -0,0 +1,182 @@
...
+
+These constraints are enough for the solidity storage model to work as expected. However when hashes are compared, they might show different behavior from the actual Keccak function (e.g. `hash(x) > hash(y)` is always true or false for given `x` and `y`, but in our formulas, the SMT solver is free to choose a function for `hash` that makes the formula satisfiable.). 
+We have not observed a practical use case yet where the numeric values of the hash function play a role, thus we chose this modeling for tractability reasons.
+
+
+### Example (Imprecision of Modeling)
+
```
mdgeorge4153, line 59: Might also be worth including a super-simple example, like
```
// sol
function getHash() ... { return keccak("hello"); }

rule checkHash {
  assert getHash() == 0x....;
}
```

(filled in with actual syntax and examples)

docs/prover/approx/hashing.md
```diff
@@ -0,0 +1,182 @@
...
+contract C {
+	function hash(uint x) public returns (bytes32) {
+		return keccak256(x);
+	}
+}
+
+```
+
+
+<!---
```
mdgeorge4153, line 87: I think this comment is obsolete

docs/prover/approx/hashing.md
```diff
@@ -0,0 +1,182 @@
...
+Most prominently, all unbounded data structures in storage (arrays, mappings) receive their storage addresses as values of the Keccak function.
+
+The Certora Prover does not operate with an actual implementation of the Keccak hash function, since this would make most verification intractable and provide no benefits in all practical cases we are aware of so far.
+Instead, the Keccak hash function is modeled as an arbitrary function that is _injective with large gaps_. 
+
+The hash function `hash` being injective with large gaps means that on distinct inputs `x` and `y`
+  - the hashes `hash(x)` and `hash(y)` are also distinct, and
+  - the gap between `hash(x)` and `hash(y)` is large enough that every additive term `hash(x) + o` that occurs in the program is also distinct from `hash(y)`.
+
+## Background: The Solidity Storage Model
```
mdgeorge4153, line 17: This section seems pretty heavy, and not super-relevant to users.  Maybe omit, or more clearly indicate how this might affect the user.

docs/prover/approx/hashing.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+
+<!---
+[comment]: # We should also clearly explain the surprising thing the user might see. A clear example of a rule that should pass and doesn't, or a rule that shouldn't pass and does, and an explanation of why.
+--->
+
+## Hashing of Unbounded Data
+
+In the discussion so far we only considered hashes of data whose length is already known before program execution (e.g. a `uint` variable always has 256 bit). Hashing of unbounded data (typically unbounded arrays, like `bytes`, `uint[]`, etc.) requires some extra measures, since their implementation requires loops and the the Certora Prover internally eliminates all loops to other constructs in order to achieve better tractability.
```
mdgeorge4153, line 93: 256 bit**s**

docs/prover/approx/hashing.md
```diff
@@ -0,0 +1,182 @@
+Modeling of Hashing in CVT
```
mdgeorge4153, line 1: I prefer to wrap the markdown files to 80 characters for various reasons

docs/prover/approx/hashing.md
mdgeorge4153: strike "to other constructs"

mdgeorge4153: possibly add a link to the page on loops
```
{doc}`internally eliminates loops <loops>`
```

docs/prover/approx/hashing.md
```diff
@@ -0,0 +1,182 @@
...
+[comment]: # We should also clearly explain the surprising thing the user might see. A clear example of a rule that should pass and doesn't, or a rule that shouldn't pass and does, and an explanation of why.
+--->
+
+## Hashing of Unbounded Data
+
+In the discussion so far we only considered hashes of data whose length is already known before program execution (e.g. a `uint` variable always has 256 bit). Hashing of unbounded data (typically unbounded arrays, like `bytes`, `uint[]`, etc.) requires some extra measures, since their implementation requires loops and the the Certora Prover internally eliminates all loops to other constructs in order to achieve better tractability.
+
+The Certora Prover models unbounded hashing similar to how it eliminates loops. The user specifies an upper length bound up to which unbounded hashing should be modeled precisely (using the CLI option {ref}`--hashing_length_bound`) as well as whether this bound is to be assumed or to be verified (using the CLI option {ref}`--optimistic_hashing`).
+
+We demonstrate how these flags work using the following program snippet.
```
mdgeorge4153, line 97: replace `.` with `:`

docs/prover/approx/hashing.md
```diff
@@ -0,0 +1,182 @@
...
+		m[b3] = v;
+	    assert(b3.length > 300, "b3 is less than 300 bytes long, unexpectedly")
+	// ...
+}
+```
+
+Let us assume that the `--hashing_length_bound` flag is set to 224 (which corresponds to 7 machine words).
+Then, the first hash operation, triggered by the mapping access `m[b1]`, behaves like the hash of a bounded data chunk. The `--optimstic_hashing` flag has no impact on this hash operation.
+Behavior of the second hash operation, triggered by the mapping access `m[b2]`, depends on whether `--optimistic_hashing` is set. 
+If the `--optimistic_hashing` flag is not set, the violation of an internal assertion will be reported by the prover, stating that an chunk of data is being hashed that may exceed the given bound of 224.
```
mdgeorge4153, line 120: It would be good to include the exact message so that people can hope to find this issue by searching for the error when they see it.

docs/prover/approx/hashing.md
```diff
@@ -0,0 +1,182 @@
...
+	// ...
+		m[b3] = v;
+	    assert(b3.length > 300, "b3 is less than 300 bytes long, unexpectedly")
+	// ...
+}
+```
+
+Let us assume that the `--hashing_length_bound` flag is set to 224 (which corresponds to 7 machine words).
+Then, the first hash operation, triggered by the mapping access `m[b1]`, behaves like the hash of a bounded data chunk. The `--optimstic_hashing` flag has no impact on this hash operation.
+Behavior of the second hash operation, triggered by the mapping access `m[b2]`, depends on whether `--optimistic_hashing` is set.
```
mdgeorge4153, line 119: I'd start a new paragraph here

docs/prover/approx/hashing.md
```diff
@@ -0,0 +1,182 @@
...
+Further use cases include direct calls of the Keccak function, either directly on solidity or inside an inline assembly snippet.
+
+Note that Certora Prover's static analysis is aware of the ABI encoder. Thus, in many cases, it can figure out that when `x, y, z` are scalars that `keccak256(abi.encode(x, y, z))` is actually a bounded hash of the form `hash(x, y, z)` as opposed to an unbounded hash of the `bytes` array that is the result of the `encode` function.
+
+
+
+
+## Conclusion
+
+To summarize, Certora Prover handles hashing in a way such that for the vast majority of hashes it will behave as expected.
```
mdgeorge4153, line 165: "...in a way that behaves as expected for most hashes."

docs/prover/approx/index.md
```diff
@@ -1,13 +1,10 @@
 Prover Approximations
 =====================
 
-In order to check programs in a finite amount of time, the Prover simplifies
-the code being verified.  This section describes the mechanisms for
-simplification.
+In order to check programs in an acceptable amount of time, the Prover simplifies the code being verified. This section describes the mechanisms for simplification.
```
mdgeorge4153, line 4: I prefer to wrap lines to 80 characters for the markdown

docs/prover/cli/options.md
```diff
@@ -328,6 +328,50 @@ The default number of loop iterations we unroll is one. However, in many cases,
...
+Options regarding hashing of unbounded data
+-------------------------------------------
+
+(--optimistic_hashing)=
+### `--optimistic_hashing`
+
+**What does it do?**
+
+ When hashing data of potentially unbounded length (including unbounded arrays, like `bytes`, `uint[]`, etc.), assume that its length is bounded by the value set through the `--hashing_length_bound` option. If this is not set, and the length can be exceeded by the input program, the prover reports an assertion violation. I.e., when this option is set, the boundedness of the hashed data assumed checked by the prover, when this option is set that boundedness is assumed instead.
+
```
mdgeorge4153, line 340: I would include a pointer to the other document here. (e.g.
```
See {doc}`../approx/hashing` for more details.
```

docs/prover/cli/options.md
```diff
@@ -328,6 +328,50 @@ The default number of loop iterations we unroll is one. However, in many cases,
...
+```
+
+(--hashing_length_bound)=
+### `--hashing_length_bound`
+
+**What does it do?**
+
+Constraint on the maximal length of otherwise unbounded data chunks that are being hashed. This constraint is either assumed or checked by the prover, depending on whether `--optimistic_hashing` has been set. The bound is specified as a number of bytes. 
+
+**When to use it?**
```
mdgeorge4153, line 358: What's the default value?

mdgeorge4153 (approved):
Looks good to me, added a few editorial comments

docs/prover/cli/options.md
alexandernutz: I was a bit cautious about including it tbh, since it might get outdated ... (the help shows it) -- but if you think we should take that risk / work load, I'll add it.
(if this was just an informative question: it's 224 bytes / 7 words)

conf.py
alexandernutz: sphinx was demanding it on my machine .. wouldn't build otherwise ... (maybe something else was missing .. I fiddled around a good bit till `make` went through ..)

alexandernutz: update: I removed it and it still seems to build --> leaving it like that (without the change to conf.py) then .

docs/prover/approx/hashing.md
alexandernutz: `i`  it is ...

docs/prover/cli/options.md
mdgeorge4153: I think it makes sense.  Someday we'll have a release process that hopefully would catch these kinds of inconsistencies when we change things... and if not, maybe we'll get a bug report

docs/prover/approx/hashing.md
alexandernutz: just to be sure: 
what should that example demonstrate? also the imprecision? i.e. that the actual `keccak(constant)` is some value and we get a violation when checking that? --> makes sense ...

docs/prover/approx/hashing.md
alexandernutz: right, done (further down there are also snippets using the other ways (direct call from solidity and inline assembly)

docs/prover/approx/hashing.md
alexandernutz: added an example that shows that we can neither prove nor disprove that the hash of a fixed string has the actual keccak value

docs/prover/approx/hashing.md
alexandernutz: hm, I could only reference it here, and push it to the end (?)
I can also extend the example with a rule that relies on the non-interference ..

alexandernutz: (done that)

docs/prover/approx/hashing.md
alexandernutz: done

docs/prover/approx/hashing.md
alexandernutz: done

docs/prover/approx/hashing.md
alexandernutz: done

docs/prover/approx/hashing.md
alexandernutz: alright, done

docs/prover/approx/hashing.md
alexandernutz: changed

docs/prover/cli/options.md
alexandernutz: good point, done

PR #57, opened 2023-02-15
Document the syntax changes introduced by CVL 2

mdgeorge4153:
Generated documentation: https://certora-certora-prover-documentation--57.com.readthedocs.build/en/57/docs/cvl/cvl2/changes.html

**Note to reviewers**: for this PR, I just want to make sure the list is complete and the descriptions are correct.  I will create a separate PR (#69) for making sure the examples work and for collecting the error messages.


docs/cvl/cvl4/changes.md
```diff
@@ -0,0 +1,371 @@
+Syntax changes introduced in CVL 4
```
shellygr, line 1: Customers don't know about CVL1,2,3 so the number 4 is much too surprising.
Prover majors are not correlated necessarily with CVL versions. 
While a breaking-change or otherwise significant update of CVL means we'll bump the Prover version, it's not symmetric: We could have a major Prover version with 0 CVL updates.

shellygr (commented):
just the first thought coming up to mind

shellygr:
Some questions that arise while fixing tests, maybe some are already on the todo:
1. How do we refer to the fallback method? previously one could write "certorafallback_0()"
2. How to convert a bytes32 to a uint256?

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+ghost uint256 sum;
+
+hook ... {
+    sum = sum + newBalance - oldBalance;
+}
+```
+will say that the right-hand side is a `mathint` which can't be assigned to a
+`uint`.
+
+We hope that security engineers will think carefully about these changes and let
```
shellygr, line 353:
```suggestion
We hope that specification writers will think carefully about these changes and let
```

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+
+The one exception is comparison operators; as mentioned above, you must add an
+explicit conversion if you want to compare two numbers with different types.
+The `to_mathint` operator exists solely for this purpose; in all other contexts
+you can simply use any number when a `mathint` is expected (since all integer
+types are subtypes of `mathint`).
+
+In order to convert from a supertype to a subtype, you must use an explicit
+cast.  In CVL 1, the syntax for casting to a subtype was `to_<subtype>(value)`,
+for example `to_uint256(x)`.
```
shellygr, line 394: We should mention those are no longer available in CVL2

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+
+CVL 2 supports the types `bytes1`, `bytes2`, ..., `bytes32`, as in Solidity.
+Number literals must be explicitly cast to these types using `to_bytesN`; for
+example:
+
+```cvl
+bytes32 x = to_bytes32(0);
+```
+
+There is no way to convert between these types and integer types (except for
```
shellygr, line 424: the crowd will ask: does `to_mathint` not work on bytesN?

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+```
+
+```{todo}
+If you do not change this, you will see the following error:
+```
+
+(cvl2-bitwise)=
+### Changes for bitwise operations
+
+The exact details for bitwise operations (such as `&`, `|`, and `<<`) were not
```
shellygr, line 438:
```suggestion
In CVL1, the exact details for bitwise operations (such as `&`, `|`, and `<<`) were not
```

docs/cvl/cvl2/migration.md
```diff
@@ -0,0 +1,9 @@
+Migration guide
+===============
+
+This section gives a step-by-step process for migrating your specs from CVL 3 to
+CVL 4.
```
shellygr, line 5:
```suggestion
This section gives a step-by-step process for migrating your specs from CVL 1 to
CVL 2.
```

shellygr (commented):
Looks very good - wrote some comments

docs/cvl/cvl2/changes.md
mdgeorge4153: I'm hoping to have this TODO resolved during the SE test fest, before we release CVL 2 to the general public.

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
+Syntax changes introduced in CVL 2
+==================================
+
+CVL 2.0 is a major overhaul to the type system of CVL.  Many
```
jtoman, line 4: I'd argue it's gone beyond the type system. I would say it's an overhaul to the "internals" and leave it at that.

jtoman: oh you literally say that on the next line. well, never mind then, leave as is

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+summarization (if any).
+
+```{todo}
+If a methods block contains a non-optional entry for a method that doesn't exist
+in the contract, you will receive the following error message:
+```
+
+### `library` annotations
+
+In CVL 2, contract functions declared as library functions must be annotated
```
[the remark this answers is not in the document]
jtoman: no, I think we pretty rarely summarized external library functions

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+```
+
+### Requirements on `returns`
+
+In CVL 2, methods block entries require a `returns` clause in the following
+situations:
+
+ - The method is not a wildcard entry and the contract function returns a value.
+   In this case, the methods block entry must match the contract function's
+   return type.
```
jtoman, line 258: wait really? you can't just do `foo(uint) envfree`?

jtoman: Also, I'm not sure what this means. If the contract function returns something you need to have a return annotation. So, isn't that always?

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+```
+
+### Comparisons require identical types
+
+When comparing two integers using `==`, `<=`, `<`, `>`, or `>=`, CVL 2 will
+require both sides of the equation to have identical types, and {ref}`implicit
+casts <cvl2-casting>` will not be used.  Comparisons with number literals (e.g.
+`0` or `1`) are allowed for any integer type.
+
+If you do not have identical types, the best solution is to use the special
```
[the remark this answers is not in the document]
jtoman: agreed.

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+
+```cvl
+assert to_mathint(balanceOf(user)) == initial + deposit;
+```
+
+Note that in this example, we do not need to cast the right hand side, since
+the result of `+` is always of type `mathint`.
+
+````{todo}
+When should you not simply cast to `mathint`?  We have one example: consider the
```
jtoman, line 325: wait, is this a todo? I say, do! this is good documentation (without the editorializing, obviously)

docs/cvl/cvl2/changes.md
jtoman: and I think **only** `to_uint256` existed, other versions didn't

docs/cvl/cvl2/changes.md
jtoman: bytes32 are described in solidity as "Fixed sized byte arrays" the fact that they are convertible to numbers is super weird imo.

jtoman: no need to repeat solidity's mistakes

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+
+### Havocing `calldataarg` variables
+
+In CVL 1, you could write the following:
+
+```cvl
+calldataarg args; env e;
+f(e, args);
+
+havoc args;
```
jtoman, line 516: I think we've decided to remove havocing for everything? am I remembering incorrectly?

jtoman (commented):
Good stuff! Some nits comments

docs/cvl/cvl2/changes.md
mdgeorge4153: The reason it's left as a TODO is because it's not clear how to generalize from this example.  Are there other situations?  What's the general rule?

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+
+```cvl
+rule example {
+    Example.S x = getAnS();
+}
+```
+
+There is no need for a `using` statement in this example.
+
+`using` statements are still required to call methods on secondary contracts.
```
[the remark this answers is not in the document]
mdgeorge4153: Do you mean pre- or post- CVL 2?  This document is intended to be the CVL 2 documentation.

docs/cvl/cvl2/changes.md
mdgeorge4153: I don't think you need to (or can) explicitly indicate that something void?

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+function _setManagedBalance(address,uint256) internal => NONDET;
+```
+
+```{todo}
+If you do not change this, you will see the following error:
+```
+
+### Required `;` in more places
+
+`using`, `pragma`, `import`, and `use` statements all require a `;` at the end.  For
```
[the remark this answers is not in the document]
thomasbernardi: `;` if there's no block, no `;` if there is a block

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+
+Every methods block entry must be marked either `internal` or `external`.  The
+annotation must come after the argument list and before the `returns` clause.
+
+If a function is declared `public` in Solidity, then the Solidity compiler
+creates an internal implementation method, and an external wrapper method that
+calls the internal implementation.  Therefore, you can summarize a `public`
+method by marking the summarization `internal`.
+
+If you want to summarize both the internal implementation and the external
```
thomasbernardi, line 157: This may or may not be the place to clarify that if the summaries are different, the `external` summary declaration will (I _believe_) take precedence, though if the `external` declaration has no summary, the internal summary will still be applied.

There is probably a better way to describe this but hopefully you get the drift

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+wrapper, you need to add two separate entries to the `methods` block.
+
+```{todo}
+If you do not change this, you will see the following error:
+```
+
+### `optional` methods entries
+
+In CVL 1, you could write an entry in the methods block for a method that does
+not exist in the contract; rules that would call the non-existent method are
```
thomasbernardi, line 167: "were skipped"? "would be skipped"?

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+```
+
+### Required `calldata`, `memory`, or `storage` annotations for reference types
+
+In CVL 2, methods entries for internal functions must contain either `calldata`,
+`memory`, or `storage` annotations for all arguments with reference types (such
+as arrays).
+
+```{todo}
+is `calldata` actually one of the options?
```
thomasbernardi, line 199: It is definitely possible to write an internal function with a parameter in a `calldata` location

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+
+```{todo}
+If you do not change this, you will see the following error:
+```
+
+### Summaries only apply to one contract by default
+
+In CVL 1, a summary in the `methods` block applied to all methods with the
+given signature.  Entries that had both an explicit receiver and a summary,
+such as the following, were disallowed:
```
thomasbernardi, line 210: I don't think this was true? We definitely had multi-contract summaries

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+
+ - The method is a wildcard entry that is summarized with a ghost or CVL
+   function summary.  In this case, the return type from the CVL function must
+   be compatible with the declared return type for the method.
+
+   ```{todo}
+   Error message
+   ```
+
+In all other situations, a `returns` clause is forbidden.
```
thomasbernardi, line 272: This makes it sound like a lot of situations. Maybe explicitly remind that this is for wildcard, non ghost/cvl function summaries

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+If you do not change this, you will see the following error:
+```
+
+Changes to integer types
+------------------------
+
+In CVL 1, the rules for casting between integer types were complex; CVL 2
+simplifies them.
+
+The general rule of thumb is that you should use `mathint` for all function
```
thomasbernardi, line 284: This is a bit confusing. It is impossible to use a `mathint` for a contract function output. Also, why would we not want `mathints` as CVL Function inputs or ghost function inputs?

docs/cvl/cvl2/changes.md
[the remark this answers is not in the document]
thomasbernardi: idk, I think this unsoundness is really a super edge case and like, our sanity checks should really catch this

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+```
+
+(cvl2-bitwise)=
+### Changes for bitwise operations
+
+In CVL1, the exact details for bitwise operations (such as `&`, `|`, and `<<`) were not
+completely specified, especially for negative integers.
+
+In CVL 2, all bitwise operations (`&`, `|`, `~`, `>>`, `>>>`, `<<`, and `xor`)
+first convert to `uint256`, then perform the operations on the full 256-bit
```
thomasbernardi, line 442: maybe instead of saying they convert to `uint256` just say they are converted to a 256-bit word? only because signed ints aren't really converted to uints, we just get a different representation. I suppose you could just say both types use two's complement encoding, because two's complement of a uint is just, it's current encoding?

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+word, then convert back to the expected type.  Signed integer types use
+twos-complement encoding.
+
+The two right-shifts differ in how they treat signed integers.  `>>` is an
+arithmetic shift; it preserves the sign bit.  `>>>` is a logical shift; it pads
+the shifted word with zero.
+
+Bitwise operations cannot be performed on `mathint` values.
```
thomasbernardi, line 450: maybe also note that the current semantics of the underlying bitwise operations are still gross overapproximations?

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+```{todo}
+If you do not change this, you will see the following error:
+```
+
+### `invoke`, `sinvoke`, and `call`
+
+Older versions of CVL had special syntax for calling contract functions:
+ - `invoke f(args);` should be replaced with `f(args);`.
+ - `sinvoke f(args);` should be replaced with `f@withrevert(args);`
+ - `call f(args)` should be replaced with `f(args)`.
```
thomasbernardi, line 483: this was not for contract functions, it was for CVL Functions (formerly called "subroutines")

docs/cvl/cvl2/changes.md
thomasbernardi: Yeah, the story here is probably, "you may no longer havoc local variables"

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+```
+
+This syntax is no longer supported.  Instead, you should declare a variable with
+the struct type:
+
+```cvl
+Example.S result = f();
+uint x = result.firstField;
+```
+
```
thomasbernardi, line 560: maybe make a note that destructuring is still supported for tuple returns

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+In CVL 1, you could declare variables of type `string[]` and `bytes[]`.  You can
+no longer use these types in CVL.
+
+You can still declare contract methods that use these types in the `methods`
+block.  However, you can only call methods that take one of these types as an
+argument by passing a `calldataarg` variable, and you cannot access the return
+value of a method that returns one of these types.
+
+```{todo}
+Determine whether you can call with `_`.
```
thomasbernardi, line 576: Very likely not. I believe the mechanism is to infer the type and generate a variable of the type internally.

thomasbernardi (changes requested):
Amazing! This is very thorough and easy to follow. Thank you for translating our working notes into something usable.

I do have some questions/comments that need to be addressed!

docs/cvl/cvl2/changes.md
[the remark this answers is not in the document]
naftali-g: Right now you can't put a semicolon also when there is no block.

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+ - `call f(args)` should be replaced with `f(args)`.
+
+```{todo}
+If you do not change this, you will see the following error:
+```
+
+### `static_assert` and `static_require`
+
+These deprecated aliases for `assert` and `require` are being removed; replace
+them with `assert` and `require`.
```
naftali-g, line 492:
```suggestion
them with `assert` and `require`, respectively.
```

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+There is no need for a `using` statement in this example.
+
+`using` statements are still required to call methods on secondary contracts.
+
+```{todo}
+Error message
+```
+
+Changes to methods block entries
+--------------------------------
```
naftali-g, line 133: We need to add to this section that user-defined types (enum, user-defined-value-type, and struct) now _must_ be written with the user-defined names. i.e. uint8 can't replace enums and tuples can't replace structs (I think in CVL1 you _had_ to use tuples?)

naftali-g (commented):
Really nice!

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,586 @@
...
+using C as c
+```
+
+becomes
+```cvl
+using C as c;
+```
+
+Note: `use` statements do not require (and may not have) a semicolon if they
+are followed by a `preserved` or `filtered` block.
```
CowardlyBattleCat, line 62:
```suggestion
are followed by a `preserved` or `filtered` block. For example,

```cvl
use rule tree_isImmutable filtered {
    f -> !excludeFromProver(f)
}
```text to allow backticks to show up in github suggestion
```

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,586 @@
...
+
+CVL 1 was less strict about the order.
+
+```{todo}
+If you do not change this, you will see the following error:
+```
+
+### Use of contract name instead of `using` variable
+
+In CVL 1, the only way to refer to a contract on the scene was to first
```
CowardlyBattleCat, line 102: I think "on the scene" is a technical term here rather than just a usage of the normal English colloquialism, yes? I was initially confused, although that may be my lack of familiarity with this usage.

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,586 @@
...
+```{todo}
+Error message
+```
+
+Changes to methods block entries
+--------------------------------
+
+In addition to the superficial changes listed above, there are some changes that
+change the way that methods block entries can be written.  In CVL 1, `methods`
+block entries often had several different functions and meanings:
```
CowardlyBattleCat, line 137:
```suggestion
block entries often have several different functions and meanings:
```
Changed to present tense since CVL 1 still exists and the points immediately below are in present tense.

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,586 @@
...
+
+ - They are used to indicate targets for summarization
+ - They are used to write generic specs that could apply to contracts with
+   missing methods
+ - They are used to declare targets `envfree`
+
+With these changes, these different uses are more explicit.
```
CowardlyBattleCat, line 144: This sentence is a bit vague. Perhaps it would be clearer to say something like, "In CVL 2, these different uses are more explicit." Not sure on the best wording.

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,586 @@
...
+with `library` in the `methods` block.
+
+```{todo}
+If you forget to declare a method as a `library` method, you will receive the
+following error message:
+```
+
+### Required `calldata`, `memory`, or `storage` annotations for reference types
+
+In CVL 2, methods entries for internal functions must contain either `calldata`,
```
CowardlyBattleCat, line 194:
```suggestion
In CVL 2, methods block entries for internal functions must contain either `calldata`,
```

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,586 @@
...
+   Error message
+   ```
+
+In all other situations, a `returns` clause is forbidden.
+
+```{todo}
+If you do not change this, you will see the following error:
+```
+
+In particular, one cannot specify return types for wildcard entries, as different
```
CowardlyBattleCat, line 278: The posted build of the docs doesn't seem to include this sentence and readthedocs is showing a build failure error https://readthedocs.com/projects/certora-certora-prover-documentation/builds/1382001/

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,586 @@
...
+
+The general rule of thumb is that you should use `mathint` for all function
+outputs, and the appropriate `int` or `uint` type for all function inputs.
+
+It is now impossible for CVL math expressions to cause overflow - all integer
+operations are exact.
+
+### Mathematical operations return `mathint`
+
+In CVL 2, the result of all arithmetic operators have type `mathint`,
```
CowardlyBattleCat, line 295:
```suggestion
In CVL 2, the results of all arithmetic operators have type `mathint`,
```

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,586 @@
...
+If you do not change this, you will see the following error:
+```
+
+
+### Havocing `calldataarg` variables
+
+In CVL 1, you could write the following:
+
+```cvl
+calldataarg args; env e;
```
CowardlyBattleCat, line 516: Syntax highlighting does not recognize `calldataarg`

docs/cvl/cvl2/migration.md
```diff
@@ -0,0 +1,85 @@
...
+You can run it on a directory containing spec files as follows:
+```
+python3 CVL1_to_CVL2.0_syntax_update.py -d . -r
+```
+
+Run `python3 CVL1_to_CVL2.0_syntax_update.py --help` for further instructions.
+
+Note that the migration script only helps deal with common use-cases where the migration is straight-forward. Some manual work and adjustment may be needed after running the script. The script may also make odd mistakes. 
+
+In particular, as the script only consumes spec files, there are decisions that it cannot do, as they are based on the Solidity code. Some of those are listed here.
```
CowardlyBattleCat, line 18:
```suggestion
In particular, as the script only consumes spec files, there are decisions that it cannot make, as they are based on the Solidity code. Some of those are listed here.
```

docs/cvl/cvl2/migration.md
```diff
@@ -0,0 +1,85 @@
...
+For example, if there is a ghost array access `forall uint x. a[x+1] == 0`, rewrite it as follows:
+
+```cvl
+forall uint x. forall uint y. to_mathint(y) == x+1 => a[y] == 0
+```
+
+### Use f.isFallback instead of comparing to `certorafallback().selector`
+
+CVL2 does not allow you to refer to the fallback function explicitly as it was seldom used and not well-defined. The most common use case for having to refer to the fallback was to check if a parametric method is the fallback function.
+For that, one can use `.isFallback` field of any variable of type `method`.
```
CowardlyBattleCat, line 55:
```suggestion
For that, one can use the `.isFallback` field of any variable of type `method`.
```

docs/cvl/cvl2/changes.md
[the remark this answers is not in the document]
mdgeorge4153: I created https://certora.atlassian.net/browse/CERT-1747

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,583 @@
...
+   In this case, the methods block entry must match the contract function's
+   return type.
+
+   ```{todo}
+   Error message
+   ```
+
+ - The method is a wildcard entry that is summarized with a ghost or CVL
+   function summary.  In this case, the return type from the CVL function must
+   be compatible with the declared return type for the method.
```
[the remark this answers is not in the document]
mdgeorge4153: I added a big warning

docs/cvl/cvl2/changes.md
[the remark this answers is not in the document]
mdgeorge4153: Do you think the example just below this is adequate?

docs/cvl/cvl2/changes.md
mdgeorge4153: I added a more extensive description

docs/cvl/cvl2/changes.md
mdgeorge4153: added a link to the glossary

docs/cvl/cvl2/changes.md
mdgeorge4153: I decided to change them all to past tense because I think that's what the rest of the document does

docs/cvl/cvl2/changes.md
mdgeorge4153: This is a known problem that should be addressed elsewhere https://certora.atlassian.net/browse/DOC-214?atlOrigin=eyJpIjoiZWJiZWMyYTg5YmUyNDAzNTllMWE4NWI1ZWYxNDlkOWMiLCJwIjoiaiJ9

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,721 @@
...
+```cvl
+using C as c
+```
+
+becomes
+```cvl
+using C as c;
+```
+
+Note: `use` statements do not require (and may not have) a semicolon if they
```
mdgeorge4153, line 61:
```suggestion
`use` statements do not require (and may not have) a semicolon if they
```

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,721 @@
...
+The changes described in this section make these different uses more explicit.
+
+### All Solidity types allowed as arguments
+
+CVL 1 had some restrictions on the types of arguments allowed in `methods` block
+entries.  For example, it was impossible to add entries for functions that
+accept arguments with user-defined types (such as enums and structs).
+
+CVL 2 `methods` block entries may use any Solidity types for arguments and
+return values, except for [function types][sol-fn-types].
```
mdgeorge4153, line 172:
```suggestion
return values, except for [function types][sol-fn-types], which are not supported.
```

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,721 @@
...
+
+In CVL 2, the methods block entry should use the same type as the Solidity
+implementations[^contract-types]:
+
+```cvl
+methods {
+    function f(Example.Permission p) internal => NONDET;
+}
+```
+
```
mdgeorge4153, line 206:
````suggestion
The method can be called from CVL as follows:
```cvl
rule example {
    f(Example.Permission.READ);
}
```

````

File: docs/cvl/cvl2/changes.md [hunk 42]
```diff
@@ -0,0 +1,721 @@
...
+In CVL 1, the rules for casting between integer types were complex; CVL 2
+simplifies them.
+
+The general rule of thumb is that you should use `mathint` for all data that is
+returned from contract functions, and the appropriate `int` or `uint` type for
+all data that will be passed as input to contract functions.
+
+It is now impossible for CVL math expressions to cause overflow - all integer
+operations are exact.
+
```
mdgeorge4153, line 419:
```suggestion
The remainder of this section describes the changes in detail.

```

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,721 @@
...
+hook ... {
+    sum = sum + newBalance - oldBalance;
+}
+```
+will say that the right-hand side is a `mathint` which can't be assigned to a
+`uint`.
+
+We hope that specification writers will think carefully about these changes and let
+us know of any other situations where the right thing is not simply casting to
+`mathint`.
```
mdgeorge4153, line 485:
```suggestion
Simply casting to `mathint` will turn overflows into vacuity.

In this particular example, the right solution is to declare `sum` to be a
`mathint` instead of a `uint`.  Note that with the more "modern" update syntax,
this isn't a problem:

```cvl
ghost uint256 sum;

hook ... {
    sum = sum + newBalance - oldBalance;
}
```
will say that the right-hand side is a `mathint` which can't be assigned to a
`uint`.

We hope that specification writers will think carefully about these changes and let
us know of any other situations where the right thing is not simply casting to
`mathint`; it is not currently clear exactly how to generalize this example.
```

File: docs/cvl/cvl2/changes.md [hunk 44]
```diff
@@ -0,0 +1,721 @@
...
+simplifies them.
+
+The general rule of thumb is that you should use `mathint` for all data that is
+returned from contract functions, and the appropriate `int` or `uint` type for
+all data that will be passed as input to contract functions.
+
+It is now impossible for CVL math expressions to cause overflow - all integer
+operations are exact.
+
+### Mathematical operations return `mathint`
```
mdgeorge4153, line 420:
```suggestion
(cvl2-mathops-return-mathint)=
### Mathematical operations return `mathint`
```

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,721 @@
...
+complicated; they depended not only on the types involved, but on the context
+in which the conversion happened.  CVL 2 simplifies these rules and improves the
+clarity and predictability of casts.
+
+In CVL 2, with one exception, you can always use a subtype whenever the
+supertype is accepted.  For example, you can always use a `uint8` where an
+`int16` is expected.  We say that the subtype can be "implicitly cast" to the
+supertype.
+
+The one exception is comparison operators; as mentioned above, you must add an
```
mdgeorge4153, line 516:
```suggestion
The one exception is comparison operators; as mentioned {ref}`above <cvl2-mathops-return-mathint>`, you must add an
```

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,721 @@
...
+The `to_mathint` operator exists solely for this purpose; in all other contexts
+you can simply use any number when a `mathint` is expected (since all integer
+types are subtypes of `mathint`).
+
+In order to convert from a supertype to a subtype, you must use an explicit
+cast.  In CVL 1, only a few casting operators (such as `to_uint256`) were
+supported.
+
+CVL 2 replaces these casting operators with two new casting operators: `assert_<type>(value)` and
+`require_<type>(value)` (for example: `assert_uint8(x)` or `require_uint8(x)`).
```
mdgeorge4153, line 527:
```suggestion
CVL 2 replaces these casting operators with two new casting operators: *assert casts*
such as `assert_uint8(x)` or `assert_int256(x)`, and *require casts* such as `require_uint8(x)` or `require_int256(x)`.
```

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,721 @@
...
+Each of these cases checks that the value is in range; the `assert` cast will
```
mdgeorge4153, line 528:
```suggestion
Each of these casts checks that the value is in range; the `assert` cast will
```

mdgeorge4153 (commented):
Doing a self-review

docs/user-guide/glossary.md
```diff
@@ -34,6 +33,22 @@ counterexample
...
+underapproximation
+  Sometimes it is useful to replace a complex piece of code with something
+  simpler that is easier to reason about.  If the approximation includes all of
+  the possible behaviors of the original code (and possibly others), it is
+  called an "overapproximation"; if it does not then it is called an
+  "underapproximation".  For example, a {ref}`NONDET <view-summary>` summary is
+  an overapproximation because every possible value that the original
+  implementation could return is considered by the Prover.
+
+  Proofs on overapproximated programs are {term}`sound`, but there may be
```
CowardlyBattleCat, line 46: image

For those of us who benefit from visual aids in examples, would it make sense to include an annotated image somewhat like this one for the glossary entry on over- and under-approximation? In particular, it could be good to show which region might contain spurious (type I error) violations and which region might contain missed (type II error) violations.

docs/prover/cli/options.md
```diff
@@ -687,3 +687,12 @@ The `solver spec` can be a single solver (`-solver=z3`) or a list of solvers
 example, `cvc5` refers to the default configuration of `cvc5` whereas
 `cvc5:nonlin` is better for nonlinear problems.  Additional options can be set
 via `z3{randomSeed=17}`.
+
+(-useBitVectorTheory)=
+#### `--settings -useBitVectorTheory`
+
+This option reduces the Prover {term}`overapproximation`s used for bitwise
```
CowardlyBattleCat, line 694: Does the word "reduces" here have a technical meaning that provides context? If not, it could be helpful to rephrase slightly to make clearer what is meant by "reduces the...overapproximations" for people with a different technical background. 

If there is a technical meaning at play, an inline synonym could be helpful.

mdgeorge4153, line 694:
```suggestion
This option uses fewer {term}`overapproximation`s for bitwise
```

docs/user-guide/glossary.md
mdgeorge4153: I feel like the glossary is maybe the wrong place for that much detail, although I'm not sure where the right place is (probably the user guide chapter on summarization?).  I'm also not sure this is the right image to include if we do decide to include one

CowardlyBattleCat: Definitely might be the wrong place for that detail, a user guide chapter on summarization could be good. I don't mean this image specifically, just something to visually show why over approximation could yield type I errors and underapproximation might yield type II errors. I initially thought that an example using sets would be good, but I didn't come up with anything that expressed the ideas well.

docs/user-guide/glossary.md
```diff
@@ -34,6 +33,22 @@ counterexample
   They all refer to an assignment of values to all of the CVL variables and
   contract storage.  See {ref}`rule-overview`.
 
+overapproximation
+underapproximation
+  Sometimes it is useful to replace a complex piece of code with something
+  simpler that is easier to reason about.  If the approximation includes all of
+  the possible behaviors of the original code (and possibly others), it is
+  called an "overapproximation"; if it does not then it is called an
+  "underapproximation".  For example, a {ref}`NONDET <view-summary>` summary is
```
thomasbernardi, line 42: is this true even if you can reach some states outside of the actual reachable states. (i.e. the approximation reaches _some_  subset of the actual reachable states but also reaches some states _outside_ the set of actual reachable sets)?

thomasbernardi (changes requested):
This is awesome. I think I have enough comments that I should request changes :)

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,721 @@
...
+variable.
+
+```{todo}
+Error message
+```
+
+Changes to methods block entries
+--------------------------------
+
+In addition to the superficial changes listed above, there are some changes that
```
CowardlyBattleCat, line 154:
```suggestion
In addition to the superficial changes listed above, there are some changes to
```

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,721 @@
...
+change the way that methods block entries can be written.  In CVL 1, `methods`
```
CowardlyBattleCat, line 155:
```suggestion
the way that methods block entries can be written.  In CVL 1, `methods`
```

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,721 @@
...
+provide an `expect` clause to the summary.  The `expect` clause tells the
+Prover how to encode the value returned by the summary.  For example:
+
+```cvl
+methods {
+    function _.foo() internal => fooImpl() expect uint256 ALL;
+}
+```
+
+This entry will replace any call to any internal function `f()` with a call to
```
CowardlyBattleCat, line 380:
```suggestion
This entry will replace any call to any internal function `foo()` with a call to
```
Should this example refer to function `foo` rather than `f` ?

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,721 @@
...
+### Mathematical operations return `mathint`
+
+In CVL 2, the results of all arithmetic operators have type `mathint`,
+regardless of the input types.  Arithmetic operators include `+`,
+`*`, `-`, `/`, `^`, and `%`, but not bitwise operators like `<<`, `xor`, and `|`
+(changes to bitwise operators are described {ref}`below <cvl2-bitwise>`).
+
+The primary impact of this change is that you may need to declare more of your
+variables as `mathint` instead of `uint`.  If you are performing arithmetic
+operations and integers that you are passing to specs, you will need to be more
```
CowardlyBattleCat, line 437: I am having trouble parsing this sentence as written. Is it possible that we want to say "...performing arithmetic operations and _generating_ integers..." or something equivalent?

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,721 @@
...
+
+hook ... {
+    havoc sum assuming sum@new == sum@old + newBalance - oldBalance;
+}
+```
+
+Simply casting to `mathint` will turn overflows into vacuity.
+
+In this particular example, the right solution is to declare `sum` to be a
+`mathint` instead of a `uint`.  Note that with the more "modern" update syntax,
```
CowardlyBattleCat, line 477: Can we assume readers will understand what we mean here by "modern" ?

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,721 @@
...
+
+```{todo}
+If you do not change this, you will see the following error:
+```
+
+### `invoke`, `sinvoke`, and `call`
+
+Older versions of CVL had special syntax for calling contract and CVL functions:
+ - `invoke f(args);` should be replaced with `f(args);`.
+ - `sinvoke f(args);` should be replaced with `f@withrevert(args);`
```
CowardlyBattleCat, line 617:
```suggestion
 - `sinvoke f(args);` should be replaced with `f@withrevert(args);`.
```

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,721 @@
...
+ - `call f(args)` should be replaced with `f(args)`.
```
CowardlyBattleCat, line 618:
```suggestion
 - `call f(args)` should be replaced with `f(args);`.
```
Was the omission of the semicolon correct?

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,721 @@
...
+```cvl
+calldataarg args; env e;
+f(e, args);
+
+havoc args;
+g(e, args);
+```
+
+In CVL 2, you can no longer `havoc` local variables.  Instead, replace the
+havoced variable with a new variable.
```
CowardlyBattleCat, line 655: Does it make sense to add:

"So the above example would become..."

?

docs/prover/cli/options.md
```diff
@@ -687,3 +687,12 @@ The `solver spec` can be a single solver (`-solver=z3`) or a list of solvers
 example, `cvc5` refers to the default configuration of `cvc5` whereas
 `cvc5:nonlin` is better for nonlinear problems.  Additional options can be set
 via `z3{randomSeed=17}`.
+
+(-useBitVectorTheory)=
+#### `--settings -useBitVectorTheory`
+
+This option uses fewer {term}`overapproximation`s for bitwise
```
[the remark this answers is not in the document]
mdgeorge4153: I'm not 100% sure.  @shellygr?

docs/cvl/cvl2/changes.md
mdgeorge4153: yes, I'm fairly certain `call` is an expression rather than a statement

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,752 @@
...
+using C as c;
+```
+
+`use` statements do not require (and may not have) a semicolon if they
+are followed by a `preserved` or `filtered` block.  For example:
+
+```cvl
+use rule tree_isImmutable filtered {
+    f -> !excludeFromProver(f)
+}
```
CowardlyBattleCat, line 68:
```suggestion
use rule poolSolvency filtered {
    f -> !isEmergencyWithdrawal(f)
}
```

CowardlyBattleCat (commented):
Made another small code change. Looks good!

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,752 @@
...
+
+You can still declare contract methods that use these types in the `methods`
+block.  However, you can only call methods that take one of these types as an
+argument by passing a `calldataarg` variable, and you cannot access the return
+value of a method that returns one of these types.
+
+```{todo}
+If you do not change this, you will see the following error:
+```
+
```
nivcertora, line 751:
```suggestion
```
If you do not change this, you will see the following error:
```
An array with an element of type bytes is currently unsupported in CVL. Only primitive elements which fit into one word are allowed.
An array with an element of type string is currently unsupported in CVL. Only primitive elements which fit into one word are allowed.
```

docs/cvl/cvl2/changes.md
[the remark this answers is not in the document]
shellygr: yes, but after fixing tons of specs I think we can use the following 2 recommendations in the guide:

1. It's fine to use `to_mathint` on a value originating from Solidity, as the original container (Solidity) is bound to 256 bits anyway. 
2. Spec values such as local variables and ghosts should not be upcasted with `to_mathint`, but rather be declared as `mathint`. An exception can be made for local variables passed as arguments into Solidity code.

docs/cvl/cvl2/changes.md
```diff
@@ -0,0 +1,721 @@
...
+}
+```
+
+[^contract-types]: There is one place where the types do not need to match exactly:
+  `address` and specific contract types (such as `IERC20`) are interchangeable.
+
+(cvl2-visibility)=
+### Required `internal` or `external` annotation
+
+Every methods block entry must be marked either `internal` or `external`.  The
```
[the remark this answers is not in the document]
mdgeorge4153: After a second read through, it doesn't seem like there's much confusion in the document so I am marking this resolved

docs/cvl/cvl2/changes.md
```diff
@@ -142,34 +133,52 @@ rule example {
...
-```{todo}
-Error message
+```cvl
+using Example as c;
+
+methods {
+    //// both are valid:
+    function c.balanceOf(address) external returns(uint) envfree;
+    function Example.transfer(address,uint) external envfree;
+}
```
thomasbernardi: Maybe it is worth it to mention somewhere that the semantics of both function entries are the same for now, but may diverge in the future once contract instances are no longer singleton?

docs/cvl/cvl2/changes.md
```diff
@@ -142,34 +133,52 @@ rule example {
...
  - They were used to indicate targets for summarization
  - They were used to write generic specs that could apply to contracts with
    missing methods
  - They were used to declare targets `envfree`
 
-The changes described in this section make these different uses more explicit.
+The changes described in this section make these different uses more explicit:
+
+```{contents}
+:local:
```
thomasbernardi, line 169: :eyes: fancy!

docs/cvl/cvl2/changes.md
```diff
@@ -210,8 +219,25 @@ rule example {
...
+
+```solidity
+function listToken(IERC20 token) internal
+```
+
+the `methods` block should use `address` for the `token` argument:
+
+```cvl
+methods {
+    function listToken(IERC20 address) internal
```
thomasbernardi, line 234: this still has type `IERC20`, oopsie

docs/cvl/cvl2/changes.md
```diff
@@ -606,38 +649,38 @@ In CVL 1, you could write a sighash instead of a method identifier in the
...
+% If you do not change this, you will see the following error:
+% ```
 
 ### `invoke`, `sinvoke`, and `call`
 
 Older versions of CVL had special syntax for calling contract and CVL functions:
- - `invoke f(args);` should be replaced with `f(args);`.
- - `sinvoke f(args);` should be replaced with `f@withrevert(args);`
+ - `invoke f(args);` should be replaced with `f@withrevert(args);`.
+ - `sinvoke f(args);` should be replaced with `f(args);`.
```
thomasbernardi, line 660: take it or leave it "should be replaced with `f(args);` _or_ `f@norevert(args);`

thomasbernardi (approved):
Looks awesome. Very clear and so important for the upcoming general release

docs/cvl/cvl2/changes.md
mdgeorge4153: I'll leave it; I can't think of any reason to recommend `@norevert`.

PR #60, opened 2023-02-20
fix semantic error in hashing doc; also some cosmetic changes

alexandernutz:
thanks @orpistiner for pointing it out!


mdgeorge4153 (approved):
looks good!

PR #63, opened 2023-03-10
Fixed a semi-vacous rule in the docs

Allarious:
Call to `contains(key)` should be cached before call to insert.
The behaviour is not incorrect since contains only changes if `succeed == true`, and `exp => true` is always true, but would be a good idea to update it for the sake of clarification.


mdgeorge4153 (approved):
looks good, thanks!

PR #65, opened 2023-03-13
DOC-304 Add -optimisticFallback and fix -optimisticReturnsize

teryanarmen:
Generated docs: https://certora-certora-prover-documentation--65.com.readthedocs.build/en/65/docs/prover/cli/options.html#settings-optimisticfallback-true


docs/prover/cli/options.md
```diff
@@ -671,8 +671,14 @@ Prover.  `--settings` should be followed by a comma-separated list of options.
 This list is incomplete.
 ```
 
+(-optimisticFallback)=
+#### `--settings -optimisticFallback=true`
+
+This option determines whether to optimistically assume unresolved fallback
```
AbhiGulati, line 677: What exactly is meant by "fallback functions" here? As I'm familiar with the term in Solidity, the `fallback` function is called when the contract receives a function call whose `calldata` does not map to any of the contract's functions. And which contract is this applying to - the `currentContract` or external calls other contracts?

teryanarmen: Thanks for your question, I updated the description to better explain the effects of this setting. In short, this setting makes unresolved state mutating calls havoc only external contracts. External to the contract the function is called from, so not necessarily `currentContract`.

docs/prover/cli/options.md
```diff
@@ -671,8 +671,21 @@ Prover.  `--settings` should be followed by a comma-separated list of options.
...
+
+This option determines whether to optimistically assume unresolved external
+calls can make arbitrary changes to all states. By default unresolved external
+calls will havoc all states. When `-optimisticFallback` is enabled, only the
+states of contracts that are external to the calling contract (not necessarily
+`currentContract`) will be havoced.
+
+Enabling this option is safe when checking invariants and parametric rules are
+used and therefore all of the functions of the current contract are checked
+independently.
```
mdgeorge4153, line 685: If I understand correctly, this doesn't seem true... even though I have checked that a method satisfies a property doesn't mean that it doesn't change the state of the contract.

docs/prover/cli/options.md
```diff
@@ -671,8 +671,21 @@ Prover.  `--settings` should be followed by a comma-separated list of options.
 This list is incomplete.
 ```
 
+(-optimisticFallback)=
+#### `--settings -optimisticFallback=true`
```
mdgeorge4153, line 675: Does this setting change the behavior of `AUTO` summaries (and is that all it does?)  If so, we should add a reference to the section on `AUTO` summaries, and also update that section to refer to this option.  You can add
```
{ref}`auto-summary` or {ref}`displayed text <auto-summary>`
```

to add a link

mdgeorge4153 (changes requested):

shellygr (approved):

PR #74, opened 2023-04-19
Eyal h/call resolution table

EyalHochCertora:
CallResolutionTable documentation, also referenced in: https://certora.atlassian.net/wiki/spaces/DR/pages/467501088/CallResolutionTable


docs/prover/Report/CallResolutionTable.md
```diff
@@ -0,0 +1,56 @@
...
+The Callee
+------
+
+The callee is composed of two elements which should be taken into account: 1. Callee contract 2. Callee sighash.
+The callee contract is the target contract of the call, the callee sighash is the sighash of the invoked function.
+As mentioned above, for internal calls, the callee will always be resolved by the tool, while for external calls, it’s not always the case.
+Here all the cases for unresolved callees are introduced:
+
+ * Fully resolved callee: In such a case, the summary will be applied iff the application policy of the summary is `ALL` 
+(see {doc}`The Methods Block — Certora Prover Documentation 0.0 documentation <summary-types>`).
```
EyalHochCertora, line 32: I am not sure how to link to another page inside the documentation, but I would like to link to: https://docs.certora.com/en/latest/docs/cvl/methods.html?highlight=summary#summary-types

docs/prover/Report/CallResolutionTable.md
```diff
@@ -0,0 +1,56 @@
+CallResolutionTable
+=================
+
+The `CallResolutionTable` shows information about all the summarized calls in the program.
+It helps the user to better understand decisions made by the Prover- which calls in the rule code got inlined, which were replaced by summaries, and why.
+
+Attributes
+------
+
+ * **Caller**- The caller of the call. Always resolved by the Prover.
```
Dmwihuri, line 10: * **Caller**- The maker of the call. Always resolved by the Prover.

docs/prover/Report/CallResolutionTable.md
```diff
@@ -0,0 +1,56 @@
...
+
+ * **Callee**- The callee of the call. For internal calls, always resolved. For external calls, might be unresolved, depending on linking, summarization, and more.
```
Dmwihuri, line 12: * **Callee**- The receiver of the call. For internal calls, always resolved. For external calls, might be unresolved, depending on linking, summarization, and more.
See below.

docs/prover/Report/CallResolutionTable.md
```diff
@@ -0,0 +1,56 @@
...
+See below.
+
+ * **Call Site**- source information of the call itself. Shows the location of the call (which file, which line), and a snippet from the source code (the invocation itself).
+
+ * **Summary**- what summary got applied for the call.
```
Dmwihuri, line 17: * **Summary**- a brief summary of the call.

docs/prover/Report/CallResolutionTable.md
```diff
@@ -0,0 +1,56 @@
...
+
+ * **Comments**- a list of comments about the resolution of the callee (which as mentioned above, might be unresolved), the applied summarization.
```
Dmwihuri, line 19: * **Comments**- a list of comments about the resolution of the callee. (As mentioned above, this might be unresolved).

docs/prover/Report/CallResolutionTable.md
```diff
@@ -0,0 +1,56 @@
...
+See below.
+
+
+The Callee
+------
+
+The callee is composed of two elements which should be taken into account: 1. Callee contract 2. Callee sighash.
```
Dmwihuri, line 26: The callee is composed of two elements which should be taken into account: 
1. Callee contract- This is the target contract of the call.
2. Callee sighash- This is the sighash of the invoked function.

docs/prover/Report/CallResolutionTable.md
```diff
@@ -0,0 +1,56 @@
...
+The callee contract is the target contract of the call, the callee sighash is the sighash of the invoked function.
```
Dmwihuri, line 27: Delete this line, already covered

docs/prover/Report/CallResolutionTable.md
```diff
@@ -0,0 +1,56 @@
...
+As mentioned above, for internal calls, the callee will always be resolved by the tool, while for external calls, it’s not always the case.
```
Dmwihuri, line 28: replace with 
For internal calls, the callee will always be resolved by the tool while for external calls, this is it’s not always the case.

docs/prover/Report/CallResolutionTable.md
```diff
@@ -0,0 +1,56 @@
...
+Here all the cases for unresolved callees are introduced:
+
+ * Fully resolved callee: In such a case, the summary will be applied iff the application policy of the summary is `ALL` 
+(see {doc}`The Methods Block — Certora Prover Documentation 0.0 documentation <summary-types>`).
+Notice, this does not mean that the call itself is resolved- it might happen, that both the callee contract and the callee sighash are resolved, but the sighash is not found in the contract.
```
Dmwihuri, line 33: replace with 
Note, this does not mean that the call itself is resolved. There could be a case where both the callee contract and the callee sighash are resolved, but the sighash is not found in the contract.

docs/prover/Report/CallResolutionTable.md
```diff
@@ -0,0 +1,56 @@
...
+
+ * Callee contract unresolved, callee sighash resolved.
+
+ * Callee Contract resolved, callee sighash unresolved.
+
+
+ Comments
+------
+
+In the comments, it is specified what is the resolution status of the callee.
```
Dmwihuri, line 46: Replace with 
The  comments, specify the resolution status of the callee.

docs/prover/Report/CallResolutionTable.md
```diff
@@ -0,0 +1,56 @@
...
+In addition, they give more insights about why the prover failed to resolve the callee.
+For example, if the callee contract is unresolved due to a wrong linking, a hint is given to the user, which specifies the slot that should be linked, and what are the possible contracts that should be considered to be linked (those that contain a function with the resolved sighash, if it is indeed resolved).
```
Dmwihuri, line 48: replace with 
For example, if the callee contract is unresolved due to wrong linking, a hint is given to the user. 

This , specifies the slot that should be linked, and what are the possible contracts that should be considered when  linking (those that contain a function with the resolved sighash, if it is indeed resolved).

docs/prover/Report/CallResolutionTable.md
```diff
@@ -0,0 +1,56 @@
...
+Moreover, the comments specify the summary application reason, which may be due to the configuration in the CLI, a summarization written in the spec, or a decision made by the Prover.
+For example, if the Prover fails to resolve the callee, it decides to havoc the call.
```
Dmwihuri, line 50: replace with 
For example, if the Prover fails to resolve the callee, it then decides to havoc the call.

Dmwihuri (commented):
Please have a look at the changes suggested.

docs/prover/Report/CallResolutionTable.md
```diff
@@ -0,0 +1,56 @@
+CallResolutionTable
+=================
+
+The `CallResolutionTable` shows information about all the summarized calls in the program.
+It helps the user to better understand decisions made by the Prover- which calls in the rule code got inlined, which were replaced by summaries, and why.
```
Dmwihuri, line 5: For clarity, The code verified by the Prover consists of several modules. However, by default, the Prover is aware of just one module. When the module is calling other modules, the Prover does not know how to identify them, let alone seeing and analyzing their code. In this case, the user guides the Prover by configuring it to identify the different modules and connect them through linking. This can be done through two ways:
1. Inlining- Taking the code of the two modules and building one big module with the code of both. 
2. Summaries- This is a mathematical description of the behavior of the module. It is usually short though less precise. It exists in two forms which are either be over approximating (describing more behaviors than there are actually implemented in the module) or under approximating (limiting the scope of what the other module can do).

docs/prover/Report/CallResolutionTable.md
mdgeorge4153: You usually want to use the ref syntax to link to a specific section:
```
(any-label-you-want)=
## Section header

...

You can link to it using {ref}`any-label-you-want` (which will display "Section Header" in the text) or {ref}`display text <any-label-you-want>`.
```

There is also the "doc" syntax to link to an entire page:
```
Link to {doc}`/docs/prover/cli/options` (note the missing .md)
```

EyalHochCertora: I have tried to use your suggestion, and pushed the code. I see it doesn't work.
I did not get the desired linking. What am I missing?

EyalHochCertora: > You usually want to use the ref syntax to link to a specific section:
> 
> ```
> (any-label-you-want)=
> ## Section header
> 
> ...
> 
> You can link to it using {ref}`any-label-you-want` (which will display "Section Header" in the text) or {ref}`display text <any-label-you-want>`.
> ```
> 
> There is also the "doc" syntax to link to an entire page:
> 
> ```
> Link to {doc}`/docs/prover/cli/options` (note the missing .md)
> ```

I have tried to use your suggestion, and pushed the code. I see it doesn't work.
I did not get the desired linking. What am I missing?

mdgeorge4153: It looks like it's a different problem now:
```
docs/prover/Report/CallResolutionTable.md: WARNING: document isn't included in any toctree
```

This indicates that the file hasn't been linked anywhere; you need to add it to the `index.md` file.

You can see the error messages by clicking through the "details" link under the failing check, although you have to log in to readthedocs.com first.

EyalHochCertora:
@mdgeorge4153 Can you approve this?

mdgeorge4153 (changes requested):
It looks like I had a review queued up, maybe I didn't finish or forgot to submit.  I remember @nd-certora saying that we should maybe not finish this one until we have more documentation in place for the rest of the rule report.

shellygr:
It does seem out of place to document the call resolution table with no guide whatsoever on the rule report. As it is dynamic I suggest to postpone working on this for now

PR #76, opened 2023-04-21
DOC-301 documenting builtin deepSanity

shellygr:


docs/cvl/builtin.md
```diff
@@ -4,7 +4,84 @@ Built-in Rules
...
 contract out-of-the-box.
 
+## Deep Sanity
+
+Enable with:
+```cvl
+use builtin rule deepSanity;
+```
+
+Ensure `--multi_assert_check` is enabled (will throw an error otherwise!).
```
Dmwihuri, line 14: replace with
Ensure `--multi_assert_check` is enabled (otherwise, it will throw an error).

docs/cvl/builtin.md
```diff
@@ -4,7 +4,84 @@ Built-in Rules
...
+
+One can configure the number of branching nodes that will be selected (it will always pick those nodes that dominate more) by setting `--settings -maxNumberOfReachChecksBasedOnDomination=N`, default `N=10`.
```
Dmwihuri, line 16: replace with
One can configure the number of branching nodes that will be selected. (the nodes that dominate more than the others will always be picked  by setting `--settings -maxNumberOfReachChecksBasedOnDomination=N`, default `N=10`.

docs/cvl/builtin.md
```diff
@@ -4,7 +4,84 @@ Built-in Rules
...
+
+### Background and motivation
+
+Sanity rules are one of our main methods to setup a new code base for verification. It serves two needs:
```
Dmwihuri, line 20: replace with 
Sanity rules act as one of our main methods to set up a new code base for verification. They serve two needs:

docs/cvl/builtin.md
```diff
@@ -4,7 +4,84 @@ Built-in Rules
...
+
+- Checking that the Prover is able to solve through a path of the code, and that there are no obvious vacuities.
```
Dmwihuri, line 22: replace with 
1. Checking that the Prover can solve through a path of the code, and that  no obvious vacuities exist.

docs/cvl/builtin.md
```diff
@@ -4,7 +4,84 @@ Built-in Rules
...
+- Checking that the Prover is able to find said path in a reasonable amount of time for both pre-processing and SMT phases.
```
Dmwihuri, line 23: replace with
2. Checking that the Prover can find the said path in a reasonable amount of time for both pre-processing and SMT phases.

docs/cvl/builtin.md
```diff
@@ -4,7 +4,84 @@ Built-in Rules
...
+
+However sanity rules are limited. As they only require to find a single path, it is not guaranteed that a fast running time for the sanity rule means that the checked method is easy. In addition, it may be able to find a path that does not go through a vacuity that does exist deeper down in the code of the program.
```
Dmwihuri, line 25: replace with
However, sanity rules are limited. This is because they only require finding a single path, and therefore it is not guaranteed that a fast running time for the sanity rule means that the checked method is easy. In addition, it may be able to find a path that does not go through a vacuity that exists deeper down in the code of the program.

docs/cvl/builtin.md
```diff
@@ -4,7 +4,84 @@ Built-in Rules
...
+function foo() {
+  ...
+  for (uint i = 0 ; i < array.len; i++) {
+    ...
+  }
+  ...
+}
+```
+
+One trivial way to pass sanity here is we can find a model where `array.len=0`. In that case, our sanity check didn’t even visit the loop, regardless of our loop configuration. Any branching in the code is potentially hiding important code.
```
Dmwihuri, line 38: replace with
One trivial way to pass sanity here is to find a model where `array.len=0`. In that case, our sanity check doesn’t visit the loop, regardless of our loop configuration. Any branching in the code is potentially hiding important code.

docs/cvl/builtin.md
```diff
@@ -4,7 +4,84 @@ Built-in Rules
...
+```cvl
+rule sanity(method f) {
+	env e;
+	calldataarg arg;
+	f(e, arg); 
+	assert false;
+}
+```
+
+That is, we call each one of the methods of the contract with arbitrary environment and arguments, and assert false. This assert false must be reached (SAT result, or red cross :x:  in the web report) for sanity to succeed.
```
Dmwihuri, line 52: replace with
We call each one of the methods of the contract with arbitrary environment and arguments, and assert false. This assert false must be reached (SAT result, or red cross :x:  in the web report) for sanity to succeed.

docs/cvl/builtin.md
```diff
@@ -4,7 +4,84 @@ Built-in Rules
...
+
+It can be alternatively written like this:
```
Dmwihuri, line 54: replace with
It can be alternatively written as:

docs/cvl/builtin.md
```diff
@@ -4,7 +4,84 @@ Built-in Rules
...
+```cvl
+rule sanity(method f) {
+    env e;
+    calldataarg arg;
+    f(e, arg); 
+    assert !true;
+}
+```
+
+Where `assert false` is replaced with `assert !true`. This is actually equivalent! But with a small tweak, it can become the key to getting much wider coverage from sanity rules.
```
Dmwihuri, line 64: replace with
Where `assert false` is replaced with `assert !true`. This is actually equivalent! However, with a small tweak, it can become the key to getting much wider coverage from sanity rules.

docs/cvl/builtin.md
```diff
@@ -4,7 +4,84 @@ Built-in Rules
...
+
+### The generalization
+
+We would like to assert certain points in the program can be reached.
```
Dmwihuri, line 68: replace with
We would like to assert that certain points in the program can be reached.

docs/cvl/builtin.md
```diff
@@ -4,7 +4,84 @@ Built-in Rules
...
+
+### The generalization
+
+We would like to assert certain points in the program can be reached. 
+Let's suppose that for every such interesting point, we add a variable assignment `X = true;`.
+We can now instead of `assert false`, write: `assert !X`. If the assert is violated, it means we went through the point that set `X` to true.
+If the assert is verified, it means we could not find a path in the program reaching the end through the point we chose, failing our sanity check.
+This is, in essence, the effect of running the `deepSanity` rule.
+
+There is an intractable number of paths in a complex code and ensuring each one can be realized does not scale.
```
Dmwihuri, line 74: replace with
There are an intractable number of paths in complex code, and ensuring that each one can be realized does not scale.

docs/cvl/builtin.md
```diff
@@ -4,7 +4,84 @@ Built-in Rules
...
+### The generalization
+
+We would like to assert certain points in the program can be reached. 
+Let's suppose that for every such interesting point, we add a variable assignment `X = true;`.
+We can now instead of `assert false`, write: `assert !X`. If the assert is violated, it means we went through the point that set `X` to true.
+If the assert is verified, it means we could not find a path in the program reaching the end through the point we chose, failing our sanity check.
+This is, in essence, the effect of running the `deepSanity` rule.
+
+There is an intractable number of paths in a complex code and ensuring each one can be realized does not scale. 
+The `deepSanity` rule uses a heuristic to pick a few interesting points in the program that must be reached:
```
Dmwihuri, line 75: replace with
Therefore, the `deepSanity` rule uses a heuristic to pick a few interesting points in the program that must be reached:

docs/cvl/builtin.md
```diff
@@ -4,7 +4,84 @@ Built-in Rules
...
+- Conditions, for example must reach the "if" or "else" case if they are code-heavy
```
Dmwihuri, line 76: replace with
1. Conditions, for example must reach the "if" or "else" case if they are code-heavy

docs/cvl/builtin.md
```diff
@@ -4,7 +4,84 @@ Built-in Rules
...
+- Before an external call
```
Dmwihuri, line 77: replace with
2. Before an external call

docs/cvl/builtin.md
```diff
@@ -4,7 +4,84 @@ Built-in Rules
...
+- The root of the program (this is the same as the usual sanity rule)
```
Dmwihuri, line 78: replace with
3. The root of the program (this is the same as the usual sanity rule)

docs/cvl/builtin.md
```diff
@@ -4,7 +4,84 @@ Built-in Rules
...
+
+The list of such interesting points will be updated from time to time.
+
+## More rules
+
 ```{todo}
-This feature is currently undocumented.
+To appear.
```
Dmwihuri, line 85: delete this line, seems misplaced

Dmwihuri (commented):
Hi @shellygr, please have a look at my suggestions.

PR #78, opened 2023-04-26
Storage comparison documentation

jtoman:
As an added bonus, cleans up some references to now yeeted features from CVL1

### DIFF
--- a/docs/cvl/expr.md
+++ b/docs/cvl/expr.md
@@ -27,10 +27,14 @@ expr ::= literal
        | function_call
 
        | expr "in" id
+	   | storage_ref storage_relop storage_ref
+
+storage_relop ::= "==" | "!="
+
+storage_ref ::= | id "@" id | id
 
 function_call ::=
-       | [ "invoke" | "sinvoke" ]
-         [ id "." ] id
+       | [ id "." ] id
          [ "@" ( "norevert" | "withrevert" | "dontsummarize" ]
          "(" exprs ")"
          [ "at" id ]
@@ -59,9 +63,18 @@ special_vars ::=
            | "lastMsgSig"
            | "_"
            | "max_uint" | "max_address" | "max_uint8" | ... | "max_uint256"
+		   
+cast_functions ::=
+    | require_functions | to_functions | assert_functions
 
-special_functions ::=
-           | "to_uint256" | "to_int256" | "to_mathint"
+require_functions ::=
+    | "require_uint8" | ... | "require_uint256" | "require_int8" | ... | "require_int256"
+
+to_functions ::=
+    | "to_mathint" | "to_bytes1" | ... | "to_bytes32"
+
+assert_functions ::=
+   | "assert_uint8" | ... | "assert_uint256" | "assert_int8" | ... | "assert_int256"
 
 contract ::= id | "currentContract"
 ```
@@ -231,8 +244,8 @@ which can be used to access the {ref}`method fields <method-type>`.
 For example,
 ```cvl
 method m;
-require m.selector == balanceOf(address).selector
-     || m.selector == transfer(address, uint256).selector;
+require m.selector == sig:balanceOf(address).selector
+     || m.selector == sig:transfer(address, uint256).selector;
 ```
 will constrain `m` to be either the `balanceOf` or the `transfer` method.
 
@@ -289,8 +302,8 @@ There are also several built-in variables:
  * The maximum values for the different integer types are available as the
    variables `max_uint`, `max_address`, `max_uint8`, `max_uint16` etc.
 
-CVL also has three built-in functions for casting mathematical types:
-`to_uint256`, `to_int256`, and `to_mathint`.  See {doc}`mathops` for details.
+CVL also has several built-in functions for converting between
+numeric types.  See {doc}`mathops` for details.
 
 
 (call-expr)=
@@ -307,12 +320,6 @@ There are many kinds of function-like things that can be called from CVL:
 
 There are several additional features that can be used when calling contract
 functions (including calling them through {ref}`method variables <method-type>`).
-
-A contract method invocation can optionally be prefixed by `invoke` or `sinvoke`,
-although this syntax is deprecated in favor of the `@norevert` and
-`@withrevert` syntax described below.  Verification of a method called with
-`invoke` will not report a counterexample if the contract method reverts, while
-`sinvoke` will.
 
 The method name can optionally be prefixed by a contract name.  If a contract is
 not explicitly named, the method will be called with `currentContract` as the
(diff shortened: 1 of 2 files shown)


docs/cvl/expr.md
```diff
@@ -359,3 +366,92 @@ state `s`.
...
+that compare storage do not encounter this behavior.
+```
+
+```{warning}
+The grammar admits storage comparisons for both equality and inequality
+that occur arbitrarily nested within expressions. However, support within the Prover for
+these comparisons is primarily aimed at assertions of storage equality, e.g., `assert s1 == s2`.
+Support for storage inequality _and_ nesting comparisons within other expressions is considered
+experimental.
+```
```
mdgeorge4153, line 449: Do we have known bugs here, or is this just a warning that this isn't fully tested yet (or that there are specific things we won't support?)  I'd be in favor of removing this warning.  Otherwise, I think it should be a bit clearer about what is / isn't supported.

mdgeorge4153: Also, I'm not sure whether "storage inequality _and_ nesting comparisons" means the combination of those two things or each of them individually

docs/cvl/expr.md
```diff
@@ -359,3 +366,92 @@ state `s`.
...
+
+```(warning
+The storage comparison checks for exact equality between every single slot of storage which can
+lead to surprising results. This behavior can happen if an uninitialized storage slot is
```
mdgeorge4153, line 453: This behavior: what behavior?

docs/cvl/types.md
```diff
@@ -271,8 +271,8 @@ Properties that compare the results of different hypothetical executions are
 sometimes called hyperproperties.
 
 CVL supports this kind of specification using the special `storage` type.  A
-variable of type `storage` represents a snapshot of the EVM storage at a given
-point in time.
+variable of type `storage` represents a snapshot of the EVM storage, as well
+as shadow state declared using ghosts, at a given point in time.
```
mdgeorge4153, line 275:
```suggestion
variable of type `storage` represents a snapshot of the EVM storage and the state of ghosts at a given point in time.
```

mdgeorge4153 (changes requested):
This looks great.  Requesting changes because the build is broken and I'd like to read through the generated documentation before merging

docs/cvl/expr.md
```diff
@@ -359,3 +366,92 @@ state `s`.
...
+Finally, if the basis is the name of a ghost function or variable, the value of that
+function/variable are compared between storage states.
+
+Two ghost functions are considered equal if they have the same outputs for all input arguments.
+
+```{warning}
+The default behavior of the Prover on unresolved external calls is to pessimistically havoc contract
+state and balances. This behavior will render most storage comparisons that incorporate such
+state useless. Care should be taken (using {ref}`summarization <summaries>`) to ensure that rules
+that compare storage do not encounter this behavior.
```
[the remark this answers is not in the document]
jtoman: It's worth noting that I was very confused why my PR wasn't working until I realized there were havocs. It's for that reason I added this.

docs/cvl/expr.md
jtoman: well, they use quantifiers, which means that we are entirely best effort.

docs/cvl/expr.md
jtoman: clarified, surprising results, which I've clarified to be "surprising failures of storage equality assertions"

docs/cvl/expr.md
```diff
@@ -29,8 +29,7 @@ expr ::= literal
        | expr "in" id
 
 function_call ::=
-       | [ "invoke" | "sinvoke" ]
-         [ id "." ] id
+       | [ id "." ] id
          [ "@" ( "norevert" | "withrevert" | "dontsummarize" ]
```
mdgeorge4153, line 33:
```suggestion
         [ "@" ( "norevert" | "withrevert" | "dontsummarize" ) ]
```

docs/cvl/expr.md
```diff
@@ -59,9 +58,19 @@ special_vars ::=
...
+		   
+cast_functions ::=
+    | require_functions | to_functions | assert_functions
 
-special_functions ::=
-           | "to_uint256" | "to_int256" | "to_mathint"
+require_functions ::=
+    | "require_uint8" | ... | "require_uint256" | "require_int8" | ... | "require_int256"
+
+to_functions ::=
```
mdgeorge4153, line 69: I thought we only have `to_mathint` now?

docs/cvl/expr.md
```diff
@@ -29,8 +29,7 @@ expr ::= literal
        | expr "in" id
 
 function_call ::=
-       | [ "invoke" | "sinvoke" ]
-         [ id "." ] id
+       | [ id "." ] id
```
mdgeorge4153, line 32:
```suggestion
       | ( [ id "." ] id | cast_function )
```

docs/cvl/expr.md
```diff
@@ -289,8 +298,8 @@ There are also several built-in variables:
  * The maximum values for the different integer types are available as the
    variables `max_uint`, `max_address`, `max_uint8`, `max_uint16` etc.
 
-CVL also has three built-in functions for casting mathematical types:
-`to_uint256`, `to_int256`, and `to_mathint`.  See {doc}`mathops` for details.
+CVL also has several built-in functions for converting between
+numeric types.  See {doc}`mathops` for details.
```
mdgeorge4153, line 302: Maybe we should put a big note at the top of `mathops` temporarily redirecting to the CVL 2 change guide?

docs/cvl/expr.md
```diff
@@ -359,3 +362,98 @@ state `s`.
 Unresolved method calls
 ```
 
+(storage-comparison)=
+Comparing Storage
+-----------------
+
+As described in {ref}`the documentation on storage types <storage-type>`, CVL represents the entirety of the EVM and its ghost state
```
mdgeorge4153, line 369:
```suggestion
As described in {ref}`the documentation on storage types <storage-type>`, CVL represents the entirety of the EVM and its {ref}`ghost state <ghosts>`
```

docs/cvl/expr.md
```diff
@@ -359,3 +362,98 @@ state `s`.
...
+
+```cvl
+using MyContract as c;
+using OtherContract as o;
+
+rule compare_state_of_c(env e) {
+   storage init = lastStorage;
+   o.mutateOtherState(e);
+   assert lastStorage[o] == init[o];
+}
```
mdgeorge4153, line 407:
```suggestion
rule compare_state_of_c(env e) {
   storage init = lastStorage;
   o.mutateOtherState(e); // changes `o` but not `c`
   assert lastStorage[c] == init[c];
}
```

docs/cvl/expr.md
```diff
@@ -359,3 +362,98 @@ state `s`.
...
+
+will pass verification whereas:
+
+```cvl
+using MyContract as c;
+using OtherContract as o;
+
+rule compare_state_of_c(env e) {
+   storage init = lastStorage;
+   c.mutateContractState(e);
```
mdgeorge4153, line 418:
```suggestion
   c.mutateContractState(e); // changes c
```

docs/cvl/expr.md
```diff
@@ -359,3 +362,98 @@ state `s`.
...
+experimental.
+```
+
+```{warning}
+The storage comparison checks for exact equality between every single slot of storage which can
+lead to surprising failures of storage equality assertions. 
+In particular, these failures can happen if an uninitialized storage slot is
+written and then later cleared by Solidity (via the `pop()` function or the `delete` keyword). After the
+clear operation the slot will definitely hold 0, but the prover will not make any assumptions
+about the value of the uninitialized slot which means they can be considered different.
```
mdgeorge4153, line 458: Is there anything the user can do about this, or are they just out of luck if they want to compare storages on a method that calls `pop`?  I don't see one, but if there is an obvious workaround, it would be worth mentioning.

mdgeorge4153 (approved):
I have a few minor nits, but otherwise looks great, very well written!

docs/cvl/expr.md
jtoman: I didn't try to clean up this whole grammar

docs/cvl/expr.md
jtoman: Uh, I didn't *really* mean to get dragged into updating the rest of the expression documentation, this was all best effort.

docs/cvl/expr.md
```diff
@@ -359,3 +362,98 @@ state `s`.
...
+   c.mutateContractState(e);
+   assert lastStorage[c] == init[c];
+}
+```
+
+will not. 
+
+```{note}
+Comparing contract's state using this method will **not** compare the balance of the contract between the
+two states.
```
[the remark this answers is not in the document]
jtoman: https://github.com/Certora/EVMVerifier/pull/4136

docs/cvl/expr.md
jtoman: you can turn on an unsafe, underapproximation flag that we are consciously keeping out of the public documentation.

docs/cvl/expr.md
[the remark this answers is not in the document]
jtoman: So, no, not yet

docs/cvl/expr.md
jtoman: This isn't correct grammar anyway, you can't do `to_uint256@withrevert(...)` which your change would suggest. Like, so much of this grammar is busted anyway, I'm not willing to hold up this PR, rushing a mistake in after approval.

PR #85, opened 2023-05-09
fixed bug in multicontract rule [DOC-307]

mdgeorge4153:


shellygr (approved):
Indentation looks off a bit? Maybe just my view

PR #86, opened 2023-05-10
[DOC-325] Editing the documentation on quantifier grounding

mdgeorge4153:
Generated documentation: https://certora-certora-prover-documentation--86.com.readthedocs.build/en/86/docs/prover/approx/grounding.html

Also made some changes to [the ref manual on quantifiers](https://certora-certora-prover-documentation--86.com.readthedocs.build/en/86/docs/cvl/expr.html#extended-logical-operations).


shellygr (approved):
There wasn't a proper git nv here so the difference in the versions isn't visible. So rubber stamping to help fix ci

docs/prover/approx/grounding.md
```diff
@@ -0,0 +1,267 @@
...
+```
+
+```{todo}
+I don't understand the following examples
+```
+
+```
+ite(forall .., x, y)
+```
+
```
yoav-rodeh, line 205: here the `forall` statement is used as the condition of an ite - that means it acts with both positive and negative polarity in the eventual formula. Easiest way to see this is looking at an `ite` in cnf form (let's say its all just boolean):
```
ite(i, t, e) = (i /\ t) \/ (!i /\ e)
```
So `i` appears both positively and negatively in the formula.

docs/prover/approx/grounding.md
```diff
@@ -0,0 +1,267 @@
...
+is bad. But:
+```
+forall a, b. ite(a, b, b + 1) > 10
+```
```
yoav-rodeh, line 209: here there is no problem, because the `forall` expression itself appears positively (actually a `assert` should be added to the beginning here)

yoav-rodeh (commented):
Great rewrite!
Keep the questions coming :)

docs/prover/cli/options.md
```diff
@@ -745,3 +745,13 @@ This option models bitwise operations exactly instead of using the default
 counterexample caused by incorrect modeling of bitwise operations, but can
 dramatically increase the time taken for verification.
 
+(-smt_groundQuantifiers)=
+#### `--settings -smt_groundQuantifiers=false`
+
+This option disables quantifier grounding.  See {ref}`grounding` for more
+information.
+
+```{todo}
```
yoav-rodeh, line 754: I'd reference to the doc above, it pretty much says when. In general it's when you get a counter example and it seems like it does not respect some quantified expression.

docs/prover/approx/grounding.md
```diff
@@ -0,0 +1,267 @@
...
+```{todo}
+Crashes how?  It would be good to include an error message here so that
+people can find it by searching.
+```
+ 
+### Recursion
+
+Quantified statements that relate a function with itself on two different inputs
+are likely to give incorrect counterexamples.  For example, the following
+`forall` statement refers to `f` twice:
```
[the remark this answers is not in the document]
mdgeorge4153: I think it's probably enough to say that it sometimes won't work.

docs/prover/approx/grounding.md
```diff
@@ -0,0 +1,274 @@
...
+
+```{todo}
+Do we have general suggestions for rewriting these?
+```
+
+(grounding-arguments)=
+### Variables must be arguments
+
+In order for grounding to work, every variable appearing in a quantified
+statement must be used at least once as an argument to a contract function.
```
yoav-rodeh, line 145: "contract function" -> "function"
Actually more correct would be "ghost function", but I think we do sort of all contract function calls, though these are very often buggy.
You can write "contract/ghost function", but we should say something somewhere saying that calling contract functions inside quantified expressions is dubious.

docs/prover/approx/grounding.md
```diff
@@ -0,0 +1,274 @@
...
+Although we can see that the assertion must be true, we would need to combine
+the statements `f(8) > f(7)` and `f(7) > f(6)` to prove it, and the grounding
+mechanism is unable to do this.
+
+In these cases, you may see a counterexample that doesn't satisfy the `require`
+statement.
+
+```{todo}
+Do we have general suggestions for rewriting these?
+```
```
yoav-rodeh, line 139: no... just don't use grounding.

docs/prover/approx/grounding.md
```diff
@@ -0,0 +1,274 @@
...
+
+Although you are allowed to call functions on complicated expressions that use
+quantified variables, doing so may produce spurious counterexamples.  For
+example, the following is allowed, but is likely to produce spurious
+counterexamples (because `x` itself is not an argument to a function):
+
+```{todo}
+I'm sort of surprised that we allow this.  It would be simpler if we just required
+that `x` must be used as a simple argument to some function, right?
+```
```
yoav-rodeh, line 165: I'm not sure we don't. But we probably should, but there might be caveats, like cases where we don't really care about the `x`. We actually generate some quantified expressions automatically, and they may not always be tight, i.e., may contain such cases. 

But it's worth a check.

docs/prover/approx/grounding.md
```diff
@@ -0,0 +1,274 @@
...
+rule.  For example, the following is disallowed, because the `forall` statement
+has double polarity.
+```cvl
+assert (forall uint x . f(x) > 0) <=> y;
+```
+
+```{todo}
+advice for rewriting?
+```
+
```
yoav-rodeh, line 238: Split to two asserts, in each one the quantified expression has a single polarity:
```
assert (forall uint x . f(x) > 0) => y;
assert y => (forall uint x . f(x) > 0);
```

yoav-rodeh (approved):
Looks great!
Just a few small comments.

docs/prover/approx/grounding.md
```diff
@@ -0,0 +1,285 @@
...
+this reason, the Prover uses an approximation called "grounding".
+
+It is not possible to ground every expression perfectly.  While grounding is
+{term}`sound` (i.e. it will not allow a rule to be verified if it is not true),
+there are cases where it may generate {term}`counterexample`s even for rules
+that should pass.  For example, a counterexample may not obey a `require`
+statement that contains a quantifier.
+
+You can prevent spurious counterexamples by turning off grounding (by passing
+{ref}`-smt_groundQuantifiers`), but without grounding the Prover may run
```
yoav-rodeh, line 15: this shows up as `-settings` when it should be `-prover_args` (`-proverArgs`?), and I think nowadays we are more for conf files? not sure how this is treated in documentation.

yoav-rodeh (approved):
lgtm

PR #87, opened 2023-05-15
Bump minimal Python version to 3.8.16 in install.md

urikirsh:
The minimal Python version we support is now 3.8.16


mdgeorge4153 (approved):
Looks good, thanks!

PR #90, opened 2023-05-19
[Proposal] RMP-6 Document the proposal for more expressive summaries

mdgeorge4153:
I'm trying a new process here.  The `expressibility/accepted` branch will contain the official, agreed-upon plan for the new feature described in [RMP-6].  To create or change the plan, we will file PRs against that branch (like this one).  This will give everyone involved an opportunity to review the PR to offer comments and stay informed.

Once the feature has been implemented and released, we will merge `expressibility/accepted` into `master`.

The documentation in this PR differs from the initial idea in [confluence][confluence-proposal] by incorporating @thomasbernardi 's suggestion that we use the same mechanism for introducing variables for the receiver as we do for binding the environment.  It also clarifies how the bound `env` variables are determined.

[confluence-proposal]: https://certora.atlassian.net/wiki/spaces/PROD/pages/480870419/With+env+and+calledContract+summarization

### DIFF
--- a/docs/cvl/methods.md
+++ b/docs/cvl/methods.md
@@ -39,7 +39,10 @@ methods          ::= "methods" "{" { method_spec } "}"
 
 method_spec      ::= ( sighash | [ id "." ] id "(" evm_params ")" )
                      [ "returns" types ]
-                     [ "envfree" ]
+                     { | "with" "(" "env" id ")"
+                       | "receiver" "(" "address" id ")"
+                       | "envfree"
+                     }
                      [ "=>" method_summary [ "UNRESOLVED" | "ALL" ] ]
                      [ ";" ]
 
@@ -98,6 +101,11 @@ as separate rules called `envfreeFuncsStaticCheck` and
 [^envfree_nonpayable]: The effect of payable functions on the contract's
   balance depends on the message value, so payable functions also require an
   `env`.
+
+Entries with summaries may also specify variables for the environment of the
+summarized function using `with(env e)`, and the address of the receiver
+contract using `receiver(address a)`.  See
+{ref}`function summaries <function-summary>` below for more details.
 
 Finally, the method entry may contain an optional summarization (indicated by
 `=>` followed by the summary type and an optional application policy).  A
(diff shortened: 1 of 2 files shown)


docs/cvl/methods.md
```diff
@@ -39,7 +39,10 @@ methods          ::= "methods" "{" { method_spec } "}"
 
 method_spec      ::= ( sighash | [ id "." ] id "(" evm_params ")" )
                      [ "returns" types ]
-                     [ "envfree" ]
+                     { | "with" "(" "env" id ")"
+                       | "receiver" "(" "address" id ")"
```
nd-certora, line 43: I think the receiver is a bad idea:
1. it can be confused with the notion of receive function in solidity
2. making “receiver” a keyword in cvl will be horrible as "receiver" is a common name used in many defi applications and I assume in many specs receiver has been used as a variable name.  
3. we have the currentContract keyword so why not have something similar? I’m ok with other names instead of 'calledContract'  but not new syntax style unless there’s a good reason

mdgeorge4153: the main reason is the difference between `with(env)` and `calledContract`.  Perhaps we could use `calledContract` and `calledEnv` or something?

mdgeorge4153: also, there's no reason we can't continue to use `receiver` as a variable or function name and also have it in this context, although I agree that it's probably not the best choice

docs/cvl/methods.md
```diff
@@ -39,7 +39,10 @@ methods          ::= "methods" "{" { method_spec } "}"
 
 method_spec      ::= ( sighash | [ id "." ] id "(" evm_params ")" )
                      [ "returns" types ]
-                     [ "envfree" ]
+                     { | "with" "(" "env" id ")"
```
naftali-g, line 42: Do we want the `with` on the left-hand side of the `=>`? or maybe on the right-hand side, similar to the `expects` clause?

docs/cvl/methods.md
naftali-g: I also thing the `receiver` is a little too much - after all it can only hold a parameter of a predetermined type, right? (this is something weird also about the `with` clause, but we're not going to change that now...).

The `calledEnv` idea is nice, I just wonder if it would be confusing for users because of the existing preserved `with` clause.

mdgeorge4153: I updated to keep `with` clause and `calledContract`.

docs/cvl/methods.md
mdgeorge4153: I moved it to the right side.

mdgeorge4153:
Should not be `envfree` and `with(env)` in the same entry

naftali-g (approved):
small nit, but looking good

docs/cvl/methods.md
```diff
@@ -307,12 +316,72 @@ Contract methods can also be summarized using CVL {doc}`functions` or
...
 that combine those variables are not supported.
 
-There are a few restrictions on the functions that can be used as approximations:
- - Functions used as summaries are not allowed to call contract functions.
- - Functions used as summaries may not have accept arguments or return values that have struct or array types.
+The function call may also use the special variable `calledContract`, which
+gives the address of the contract on which the summarized method was called.
+This is useful for identifying the called contract in {ref}`wildcard summaries
+<cvl2-wildcards>`.
+
```
nd-certora, line 332: should we state the calledContract can only be used with the summarization expression and not in any other cvl statement?

nd-certora (approved):

PR #92, opened 2023-05-23
Added section for satisfy statements

jhoenicke:


docs/cvl/statements.md
```diff
@@ -93,6 +94,38 @@ Unlike Solidity's `assert` and `require`, the CVL syntax for `assert` and
...
+such that for every `satisfy` statement there is a witness that executes this
+statement.
+
+If for at least one `satisfy` statement there is no execution path on which it
+holds, an error is reported.  If all `satisfy` statements can be fulfilled on
+at least one path, the rule succeeds.
+
+```{note}
+A success does only guarantee that there is some execution starting in some
+arbitrary state.  It is not possible to check that there is an execution for
```
andrew-certora, line 125: Doesn't `assert` check that the expression holds for every starting state? Should we direct the reader to try specifying this with assert instead if this is their goal?

andrew-certora: OK sounds reasonable.

andrew-certora (approved):

shellygr (approved):
Looks good. Thanks!

PR #93, opened 2023-05-23
[Proposal] remove library keyword; add `storage` modifiers for `external` functions

mdgeorge4153:
Based on the conversation from 5/23 with myself, Naftali, and Thomas, this seems to be the best resolution for the problem of reference types in libraries.

 - external functions may now include `storage` annotations on arguments with reference types
 - the `library` keyword is removed
 - to determine whether a method entry applies, we will generate both library and non-library sighashes; we can do this exactly with the addition of the `storage` modifier.

Generated docs: https://certora-certora-prover-documentation--93.com.readthedocs.build/en/93/docs/cvl/cvl2/changes.html#required-calldata-memory-or-storage-annotations-for-reference-types

### DIFF
--- a/docs/cvl/cvl2/changes.md
+++ b/docs/cvl/cvl2/changes.md
@@ -340,7 +340,7 @@ In CVL 2, methods block entries for internal functions must contain either `call
 as arrays).
 
 Entries for external functions may have `storage` annotations for argument
-references.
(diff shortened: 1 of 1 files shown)


thomasbernardi (changes requested):
I feel like this diff should also include some bit about how summaries will apply to both library and contract versions of wildcard entries.

PR #96, opened 2023-05-29
Dealing with library contracts - documenting esoteric (new) flags

shellygr:


nd-certora (approved):
nice, it will be nice to have these options also under the cli flags but that sections needs a redo

PR #102, opened 2023-06-07
Heavy editing to builtin rules documentation

mdgeorge4153:
I made pretty extensive edits, so I created a separate branch for review.

I left some `{todo}` comments for things I wasn't sure about.

Generated docs: https://certora-certora-prover-documentation--102.com.readthedocs.build/en/102/docs/cvl/builtin.html


docs/cvl/builtin.md
```diff
@@ -1,113 +1,195 @@
...
-function foo() {
-   // ...
-  for (uint i = 0 ; i < array.len; i++) {
-   // ...
-  }
- // ...
-}
+```{todo}
+What does "sanity fails" mean?  Is it still the case that a "failing" sanity
+show up as success and vice-versa?  If they are still inverted, we should
```
shellygr, line 101: We're changing that soon to use satisfy so no need for word juggling

docs/cvl/builtin.md
```diff
@@ -1,113 +1,195 @@
...
+clearly explain that.
 ```
 
-One trivial way to pass sanity here is to find a model where `array.len=0`. In that case, our sanity check does not visit the loop, regardless of our loop configuration. Any branching in the code is potentially hiding important code.
+```{todo}
+What kind of failure does it indicate?  What should the user do if they see a
```
shellygr, line 106: This warrants a guide of its own. Usually just loop unroll config

docs/cvl/builtin.md
```diff
@@ -1,113 +1,195 @@
...
+The Certora Prover has built-in general-purpose rules targeted at finding common
+vulnerabilities.  These rules can be verified on a contract without writing any
+contract-specific rules.
 
+Built-in rules can be included in any spec file by writing `use builtin rule
+<rule-name>;`.  This document describes the available built-in rules.
+
+```{todo}
+I would be in favor of removing the two "How ... is checked" sections; it seems
+like an implementation detail and I'm not sure that it's useful to users.  I
```
shellygr, line 13: Not all indeed, but I feel some users would appreciate a better understanding of the prover work

docs/cvl/builtin.md
```diff
@@ -1,113 +1,179 @@
...
-- `msgValueInLoopRule` – checks for occurrences of `msg.value` and delegate calls in loops.
-- `hasDelegateCalls` - checks for delegate calls anywhere in the contract.
-- `sanity`
-- `deepSanity`
+Loops that [use `msg.value`][msg-value-vulnerability] or [make delegate
+calls][delegatecall-vulnerability] are a well-known source of security
+vulnerabilities.
 
+[msg-value-vulnerability]: https://trustchain.medium.com/ethereum-msg-value-reuse-vulnerability-5afd0aa2bcef
+[delegatecall-vulnerability]: https://blog.trailofbits.com/2021/12/16/detecting-miso-and-opyns-msg-value-reuse-vulnerability-with-slither/
```
ghilac, line 39: I would not cite slither...
https://blog.solidityscan.com/security-issues-with-delegate-calls-4ae64d775b76 maybe this one?

docs/cvl/builtin.md
```diff
@@ -1,113 +1,179 @@
...
 ```
+in a spec file.  The rule will fail on any functions that can make delegate
+calls or access `msg.value` inside a loop.
 
-The sanity rule acts as one of our main methods to set up a new code base for verification. It serves two needs:
+```{todo}
+I imagine it's actually on any functions that recursively call any functions that do this?
+And probably also loops that call functions that recursively access `msg.value`?
+Not sure if this needs more clarification or not.
+```
```
ghilac, line 56: yes it is but I am not sure if it's obvious or not. Maybe we should only add a clarification or in the sentence above

ghilac (commented):
Adding only some minor comments

docs/cvl/builtin.md
sitvanit83: The link for msg.value mentions the relation to delegate calls so it is enough for the msg.value/delegate calls in loops.
I added the link Ghila gave to hasDelegateCalls as this refers to delegate calls in general and does not mention loops.

docs/cvl/builtin.md
sitvanit83: It is. We have tests for this and I checked it again.

PR #110, opened 2023-06-20
DOC-168: Update docs for -solver et al

nafur:
Naming convention:

Before requesting review:
 - Make sure CI is passing

Jira ticket: TODO


docs/prover/cli/options.md
```diff
@@ -726,19 +726,26 @@ reports may not be generated.
...
-via `z3{randomSeed=17}`.
+By default, a portfolio of SMT solvers using various configurations is used within
+the Prover.  It can be useful to specify only a subset of these to save on
+computation time.  In rare cases, solver specific options can improve performance
+as well.  Setting `-solver=<solver spec>` filters the predefined portfolio to only
+use those configuration that match the given solver specification.
+
+The `solver spec` can be a single solver (`-solver=z3:def`), or a list of solver
+configurations (`-solver=[z3:def,cvc5:def]`), where each such solver can be
+further modified.  For example, `cvc5` refers to the set of pre-configured
```
martin-hruska, line 737: I would add here an explicit example of the CLI flag, i.e., "For example, `cvc5` (`-solver=cvc5`) refers to..." since the previous examples show only a specific configuration of solver.

docs/prover/cli/options.md
```diff
@@ -726,19 +726,26 @@ reports may not be generated.
...
+configurations of `cvc5` whereas `cvc5:nonlin` is a specific configuration used
+for nonlinear problems.  Additional options can be set via `z3{randomSeed=17}`.
+
+With `-smt_overrideSolvers=true`, the portfolio can be replaced instead of filtered.
```
martin-hruska, line 741: Again, I would add here a full example of the use of this flag, something like "`-smt_overrideSolvers=true,-solver=cvc5:def` would cause using only the default configuration of `CVC5`".

martin-hruska (commented):
I have only two minor comments.

docs/prover/cli/options.md
```diff
@@ -726,19 +726,26 @@ reports may not be generated.
...
+For even better control of which solvers are used in which situation, solver
+specification for certain logics can be given via `-smt_LIASolvers=<solver spec>`, 
+`-smt_NIASolvers=<solver spec>`, and `-smt_BVSolvers=<solver spec>` for linear,
+non-linear and bit-vector formulas.
 
 (-useBitVectorTheory)=
-#### `--settings -useBitVectorTheory`
+#### `--settings -smt_useBV`
```
jar-ben, line 748: Since we are already updating the BV flag here, perhaps it would be good to include a note like the following. 
`The disadvantage of the BV encoding is that it does not allow modelling mathints; in particular, the maximum supported integer value is 2^256-1 (uint256). We currently do not have a setting/encoding that would both model precisely bitwise operations and allow for mathints`.

jar-ben (approved):
Lgtm. Just minor suggestion.

docs/prover/cli/options.md
jar-ben: Do we actually have somewhere the whole list of available solver configurations? I am not sure the user can access it. So perhaps we should also add the list here?

docs/prover/cli/options.md
shellygr: agree

docs/prover/cli/options.md
```diff
@@ -726,19 +726,26 @@ reports may not be generated.
...
-example, `cvc5` refers to the default configuration of `cvc5` whereas
-`cvc5:nonlin` is better for nonlinear problems.  Additional options can be set
-via `z3{randomSeed=17}`.
+By default, a portfolio of SMT solvers using various configurations is used within
+the Prover.  It can be useful to specify only a subset of these to save on
+computation time.  In rare cases, solver specific options can improve performance
+as well.  Setting `-solver=<solver spec>` filters the predefined portfolio to only
+use those configuration that match the given solver specification.
+
+The `solver spec` can be a single solver (`-solver=z3:def`), or a list of solver
```
shellygr, line 735: can we list all the portfolios?

shellygr (commented):
this is fine but not constructive to the end user, see my comments below. 
(I'm basically repeating Jaroslav)

mdgeorge4153 (approved):
Looks good to me, modulo the changes that others have proposed already

docs/prover/cli/options.md
nafur: I don't think this makes a lot of sense, because it will be outdated anyway. I'd rather make sure that we print the portfolio that is being used in the logs so that a user can figure it out on whatever CVT version is being used.
Also, it is a bit more subtle than described here: we actually have a matrix of portfolios right now, based on the logic (LIA, NIA, BV) and the timeout (<5s, <10s, >10s).

docs/prover/cli/options.md
nafur: > Do we actually have somewhere the whole list of available solver configurations? I am not sure the user can access it. So perhaps we should also add the list here?

As before: IMHO we should not add lists of solver configurations here, as they will be outdated anyway. I'd rather establish a standard way to inspect this on a CVT executable.

docs/prover/cli/options.md
nafur: I've had a quick look how we could allow a user to inspect the solver configs... it is not clear to me how to do this... still I think we should not put a (soon outdated) list here.

martin-hruska (approved):

PR #118, opened 2023-07-09
Added documentation to readOnlyReentrancy builtin rule

sitvanit83:
Naming convention:

Before requesting review:
 - Make sure CI is passing

Jira ticket: TODO

### DIFF
--- a/docs/cvl/builtin.md
+++ b/docs/cvl/builtin.md
@@ -25,6 +25,7 @@ built_in_rule_name ::=
     | "hasDelegateCalls"
     | "sanity"
     | "deepSanity"
+    | "readOnlyReentrancy"
 ```
 
 (built-in-msg-value-in-loop)=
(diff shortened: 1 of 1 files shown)


docs/cvl/builtin.md
```diff
@@ -173,3 +174,29 @@ instruments the contract code at `p` to set `x_p` to `true`.  The Prover then
...
+The `readOnlyReentrancy` rule can be enabled by including
+```cvl
+use builtin rule readOnlyReentrancy;
+```
+in a spec file.  Any functions that have read only reentrancy will fail the
+`readOnlyReentrancy` rule.
+
+### How `readOnlyReentrancy` is checked
+
+The `readOnlyReentrancy` rule adds instrumentation that is used to check that the results
```
mdgeorge4153, line 195: I think there's some missing context here about when these checks happen.  If I had to guess, I think you mean the following:

> Reentrancy vulnerabilities can arise when a contract makes an external call with an inconsistent internal state.  This behavior allows the receiver contract to make reentrant calls that exploit the inconsistency.
> 
> The `readOnlyReentrancy` rule ensures that whenever method `f` of contract `C` makes an external call, the internal state of `C` is equivalent to either (1) the state of `C` before `f` started, or (2) the state of `C` after `f` returns (by "equivalent", we mean that all view functions return the same values).  This ensures that the external call cannot observe `C` in any state that it couldn't have without being called from `C`.  It prevents external calls from a partially-updated state where some variables have been updated but others still have their old values.

Is this correct?

docs/cvl/builtin.md
```diff
@@ -173,3 +174,29 @@ instruments the contract code at `p` to set `x_p` to `true`.  The Prover then
...
+The `readOnlyReentrancy`  built-in rule detects 
+[read only reentrancy vulnerabilities in a contract][readonly-reentrancy-vulnerability].
+
+[readonly-reentrancy-vulnerability]: https://blog.pessimistic.io/read-only-reentrancy-in-depth-6ea7e9d78e85
+
+The `readOnlyReentrancy` rule can be enabled by including
+```cvl
+use builtin rule readOnlyReentrancy;
+```
+in a spec file.  Any functions that have read only reentrancy will fail the
```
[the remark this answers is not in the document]
sitvanit83: done

docs/cvl/builtin.md
sitvanit83: Almost. We compare the result of the view function to its result when the calling function starts which does not match (1) and (2) above and to the result of the view function at the end of the calling function which also can be different from (1) and (2).

docs/cvl/builtin.md
```diff
@@ -173,3 +174,32 @@ instruments the contract code at `p` to set `x_p` to `true`.  The Prover then
...
+-----------------------------------------------------------
+
+The `readOnlyReentrancy`  built-in rule detects 
+[read-only reentrancy vulnerabilities in a contract][readonly-reentrancy-vulnerability].
+
+[readonly-reentrancy-vulnerability]: https://blog.pessimistic.io/read-only-reentrancy-in-depth-6ea7e9d78e85
+
+The `readOnlyReentrancy` rule can be enabled by including
+```cvl
+use builtin rule readOnlyReentrancy;
```
ghilac, line 188: I think that it's still viewReentrancy. I can merge  the change if we think it's needed it but I didn't want to do that before the conferences

docs/cvl/builtin.md
mdgeorge4153: I think we are saying the same thing, because I said that `f` is the calling function...you replaced it with "the calling function".  I would either (1) remove `f` entirely and keep the remainder, or (2) change it back to `f`, but maybe add "makes an external call to `g`" to avoid the confusion.

mdgeorge4153 (approved):
I had one response to the unresolved comment, but otherwise looks good!

PR #124, opened 2023-08-09
fixes for #71

guydorens:
- Updated screenshot - the results of the failing sanity checks appear in the problems view.
 - Removing notes about naming - everything is renamed according to the naming conventions in https://github.com/Certora/Documentation/pull/71
 - Addressing the question in the `todo` block.


nd-certora (approved):
great!

PR #125, opened 2023-08-10
remove solc_args

rahav-certora:
Naming convention:

Before requesting review:
 - Make sure CI is passing

Jira ticket: TODO


docs/prover/cli/options.md
```diff
@@ -251,19 +251,6 @@ Whenever you want to use a Solidity compiler executable with a non-default name.
 **Example**  
 `certoraRun Bank.sol --verify Bank:Bank.spec --solc solc8.1`
 
-### `--solc_args`
```
mdgeorge4153, line 254: I don't understand, was this option removed?  It is still reported in `certoraRun --help`

mdgeorge4153 (changes requested):

docs/prover/cli/options.md
```diff
@@ -251,19 +251,6 @@ Whenever you want to use a Solidity compiler executable with a non-default name.
...
-
-**What does it do?**
```
mdgeorge4153, line 256: I might leave this section in but replace it with
````
```{versionchanged} 4.0
Support for `--solc_args` has been removed; see {ref}`--optimize`, ...
```
````

docs/prover/cli/options.md
rahav-certora: This option was removed. Once merged this option will not be listed in --help

docs/prover/cli/options.md
rahav-certora: I am not sure it is a good idea. From what I see `--solc_args` was not used for long time by customers. It is only for solc options that are not supported explicitly  like --solc_allow_path, solc_optimize, solc_via_ir etc,  and it seems that what we provide gives and gave good coverage

mdgeorge4153 (approved):

PR #133, opened 2023-08-21
upgraded to new rtd format

mdgeorge4153:
RTD is requiring us to upgrade to the a new config format


shoham-certora (approved):
Looks good to me.

PR #136, opened 2023-08-25
removed support for windows

mdgeorge4153:
Link to generated documentation: https://certora-certora-prover-documentation--136.com.readthedocs.build/en/136/docs/user-guide/getting-started/install.html

### DIFF
--- a/docs/user-guide/getting-started/install.md
+++ b/docs/user-guide/getting-started/install.md
@@ -65,7 +65,10 @@ can arise during installation and how to deal with them:
 <details>
   <summary>Windows</summary>
 
-  The Certora Prover is not supported on Windows.  Windows users should use WSL.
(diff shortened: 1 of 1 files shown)


docs/user-guide/getting-started/install.md
```diff
@@ -5,6 +5,15 @@ Installation
 Step 1: prerequisites
 ---------------------
 
+<details>
+  <summary>Linux or MacOS</summary>
+  Windows users should use [WSL][wsl].
+
+  % TODO: more information?
```
RedLikeRosesss, line 12: Do we really need to TODO here? 
I mean I understand that we need to add more info, but maybe users don't need to see it.

shellygr: they won't, it's a comment

shellygr (approved):

PR #137, opened 2023-08-30
[DOC-319] Improvements to `invariants` page

mdgeorge4153:
Jira ticket: https://certora.atlassian.net/browse/DOC-319
Link to generated documentation: https://certora-certora-prover-documentation--137.com.readthedocs.build/en/137/docs/cvl/invariants.html


docs/cvl/invariants.md
```diff
@@ -205,7 +207,22 @@ we are checking for preservation.
 To see why this is not the desired behavior, consider a `deposit` method that
 increases the message sender's balance.  When the
 `zero_address_has_no_balance_v2` invariant is checked on `deposit`, the Prover
-will report a violation with the `msg.sender` set to 0 in the call to `deposit`
+will effectively check the following (see {ref}`invariant-as-rule`):
```
nafur, line 210: Okay, I see *what* is happening here. I may very well not see something obvious here, but *why* do we have both `e` and `calledEnv`? Maybe there should be some more general explanation (independent of the `preserved` stuff) of how `invariants` behave across multiple messages?

docs/cvl/invariants.md
```diff
@@ -245,20 +263,21 @@ method:
...
 
 In this example, when the variable `f` is bound to `deposit(uint)`, the
-expression `f.selector != deposit(uint).selector` evaluates to `false`, so the
+expression `f.selector != sig:deposit(uint).selector` evaluates to `false`, so the
 method will be skipped.
 
 ```{note}
-If there is a {ref}`preserved block <preserved>` for a method it will be
-verified even if it should be filtered out.
+If there is a {ref}`preserved block <preserved>` for a method, the method will
```
nafur, line 275: Oh, good that I asked. This is very different from what I originally understood. So the verification logic is something like this?
```
fun doVerify(method m) {
  if (hasSpecialPreservedBlock(m)) verify(m);
  if (evaluateFilter(m)) verify(m);
  skip(m);
}
```
I wonder whether we should at least throw a warning here ("invariant filter for <i> excluded method <m>, but a specialized preserved block was defined"). I'd argue that one way or another, the user is confused or forgot to delete something :-)

nafur (commented):
Thanks! Two more questions...

docs/cvl/invariants.md
mdgeorge4153: I think this is covered in the section on invariant assumptions; I added a link here to that section.

PR #141, opened 2023-09-04
CERT-3221: clarify top-level user-defined type usage

naftali-g:
The docs about how to reference top-level user-defined types is misleading - one must use one of the contracts in the scene, and not any contract at all. Clarify this.


mdgeorge4153 (approved):
Looks good, would be good to link to (and fix) the glossary entry on "scene"

PR #148, opened 2023-09-25
[DOC-340] Update hook reference manual

mdgeorge4153:
[Jira ticket](https://certora.atlassian.net/browse/DOC-340?atlOrigin=eyJpIjoiMTdlYTlmNTMyNzk5NGVkOWI5YzdkZTA2ZWFhNTk4OWUiLCJwIjoiaiJ9)

Link to generated documentation:
 - [Hooks reference manual](https://certora-certora-prover-documentation--148.com.readthedocs.build/en/148/docs/cvl/hooks.html)
 - [Opcode hooks user guide](https://certora-certora-prover-documentation--148.com.readthedocs.build/en/148/docs/user-guide/opcodes.html#using-opcodes)
 - [Ghosts user guide](https://certora-certora-prover-documentation--148.com.readthedocs.build/en/148/docs/user-guide/basics/ghosts.html)

This PR migrates the old documentation on hooks into the reference manual and the user guide, and combines the ref. manual page on hooks with the page on EVM opcode hooks.

Note that this PR is related to https://github.com/Certora/EVMVerifier/pull/5288 (which contains tests of the interactions between hooks and summaries) and https://github.com/Certora/Examples/pull/46 (which contains an example that this documentation links to)


docs/cvl/hooks.md
```diff
@@ -20,77 +22,422 @@ Syntax
...
+on accesses to specific contract variables or specific array, struct, or
+mapping accesses.
+
+Storage locations are designated by "access paths".  An access path
+starts with either the name of a contract field, or a [slot number][storage-layout].
+
+```{todo}
+is this correct?
+
+Contract fields must be qualified by the contract that defines them (e.g.
```
shellygr, line 115: I don't recall if we support the `Contract.` syntax, but default to `currentContract` is probably true. Maybe @naftali-g remembers

docs/cvl/hooks.md
```diff
@@ -20,77 +22,422 @@ Syntax
...
+`Contract.field`).  If the contract name is omitted, it defaults to
+`currentContract`.
+```
+
+```{todo}
+Does the contract need to be the contract that defines the field, or can it be
```
shellygr, line 121: iirc we don't care about the inheritance for naming/identifying fields. if the same private field appears in more than one contract in the inheritance structure, we omit those completely and they are inaccessible. the classic example is the `_gap` static array used by upgradable contracts

docs/cvl/hooks.md
```diff
@@ -20,77 +22,422 @@ Syntax
...
+loads and stores that are a specific number of bytes from the
+base.  For example, the following hook will match writes to the third or fourth
+byte of slot 1 (these two bytes are matched because the type of the variable is
+`uint16`:
+
+```cvl
+hook Sstore (slot 1).(offset 2) uint16 b STORAGE { ... }
+```
+
+These different kinds of paths can be combined.  For example, the following
```
shellygr, line 169: can we combine low level access and 'high level' access?

docs/cvl/hooks.md
```diff
@@ -20,77 +22,422 @@ Syntax
...
+```cvl
+hook C.users[KEY address user].balance uint v (uint old_value) STORAGE { ... }
+```
+Inside the body of the hook, the variable `user` will refer to the address that
+was used as the key into the mapping `C.users`; the variable `v` will contain
+the value that is written, and the variable `old_value` will contain the value
+that was previously stored there.
+
+```{note}
+The only available access paths for `solc` versions 5.17 and older are `slot`
```
shellygr, line 181: not 5.13? just making sure

docs/cvl/hooks.md
```diff
@@ -20,77 +22,422 @@ Syntax
...
+and `offset` paths.
+```
+
+### Access path caveats
+
+```{todo}
+information on what kind of funny business happens if people do their own memory
+management, and about the disjointness of arrays and so forth
+
+How can you tell if an analysis failed and your hooks didn't apply?
```
shellygr, line 191: maybe suffices to say that if the relevant analysis fails and hooks cannot be applied, the rule will fail with a suitable message and an error will show in the errors tab of the web report

docs/cvl/hooks.md
```diff
@@ -20,77 +22,422 @@ Syntax
...
+first, and then the `ALL_SLOAD` hook.
+
+Similarly, if a store would trigger both an `Sstore` pattern and an `ALL_SSTORE`
+pattern, the `Sstore` hook would be executed, followed by the `ALL_SSTORE` hook.
+
+```{note}
+Just like the usual opcode hooks, the raw storage hooks are applied on all
+contracts.  This means that a storage access on _any_ contract will trigger the
+hook.  Therefore, in a rule that models calls to multiple contracts, if two
+contracts are accessing the same slot the same hook code will be called with
```
shellygr, line 237: what about `executingContract` that was recently added?

docs/cvl/hooks.md
```diff
@@ -20,77 +22,422 @@ Syntax
...
+the same slot number.
+```
+
+(opcode-hooks)=
+EVM opcode hooks
+----------------
+
+```{todo}
+Are `Create` hooks and `CREATE1`/`CREATE2` hooks different?  Should `Create`
```
shellygr, line 246: they are slightly different, let's skip documentation for `Create` to a later PR

docs/cvl/hooks.md
```diff
@@ -20,77 +22,422 @@ Syntax
...
+Hook bodies
+-----------
+
+The body of a hook may contain almost any CVL code, including calls to other
+Solidity functions.  The only exception is that hooks may not contain
+parametric method calls.  Expressions in hook bodies may reference variables
+bound by the hook pattern.
+
+````{todo}
+Questions about the following:
```
shellygr, line 378: the core point here is due to summaries being applied before hooks.
so since some summaries like internal summaries are applied before CVL, we get those in the hook bodies.
I suspect that a summary that did not apply early like will also not get applied after hook inlining.

we should also test the 2 cases you mentioned just to make sure our understanding is correct, and to consider which summaries are applied early and which are not (I am mostly suspicious about external summaries).

shellygr (changes requested):
looks good, responded on the todos

docs/cvl/hooks.md
mdgeorge4153: The old docs just said "before version 5" and 5.17 is the latest release in version 5

docs/cvl/hooks.md
```diff
@@ -20,77 +22,422 @@ Syntax
...
-             |   static_slot_pattern:sp {: RESULT = sp; :}
-             ;
+hook ::= "hook" pattern block
+
+pattern ::= "Sstore" access_path param [ "(" param ")" ] "STORAGE"
+          | "Sload"  param access_path "STORAGE"
+          | opcode   [ "(" params ")" ] [ param ]
+
+access_path ::= id
+              | "(" "slot" number ")"
```
mdgeorge4153, line 32: Should be `[ id "." ] "(" "slot" number ")"`

docs/cvl/hooks.md
```diff
@@ -20,77 +22,422 @@ Syntax
...
+```cvl
+hook Sstore C.totalSupply uint ts STORAGE { ... }
+```
+
+```{todo}
+Is this correct?
+
+If there is a store hook that binds the old value of the variable, then the
+Prover will add a load instruction that reads the value immediately before the
+store instruction.  Therefore, if a path has both a load hook and a store hook,
```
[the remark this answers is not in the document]
mdgeorge4153: You're correct, this is not true (removing)

docs/cvl/hooks.md
mdgeorge4153: testing shows that this is correct

docs/user-guide/basics/ghosts.md
```diff
@@ -0,0 +1,55 @@
...
+
+Together, hooks and ghosts let you keep track of what the contract does during
+execution.  This guide works through some common examples of using ghosts and
+hooks to monitor contracts.  For full details see the reference manual on
+{ref}`hooks <hooks>` and {ref}`ghosts <ghosts>`.
+
+```{contents}
+```
+
+## Running example
```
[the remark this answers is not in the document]
mdgeorge4153: currently nobody; I just replaced the missing parts with a todo pointing to the relevant references

docs/cvl/hooks.md
mdgeorge4153: I'm running a bunch of tests now (https://github.com/Certora/EVMVerifier/pull/5288), but based on discussion in #cvl-development I'm pretty sure the doc is accurate now.

shellygr (approved):
lgtm

PR #150, opened 2023-10-04
CERT-3448 Fix the documentation for gambit a bit

chandrakananandi:
Links:
 - [Gambit](https://certora-certora-prover-documentation--150.com.readthedocs.build/en/150/docs/gambit/gambit.html)
 - [Mutation verifier](https://certora-certora-prover-documentation--150.com.readthedocs.build/en/150/docs/gambit/mutation-verifier.html)

### DIFF
--- a/docs/gambit/gambit.md
+++ b/docs/gambit/gambit.md
@@ -242,12 +242,11 @@ Here are some examples of using the `--sourceroot` option.
    ```
 
    This should output the following:
-   <!-- Code output: using `pre` to avoid the Copy To Clipboard feature -->
-   <pre>
+   ```
    Generated 1 mutants in 0.13 seconds
    1,BinaryOpMutation,benchmarks/BinaryOpMutation/BinaryOpMutation.sol,23:10, % ,*
    gambit_out/mutants/1/benchmarks/BinaryOpMutation/BinaryOpMutation.sol
-   </pre>
+   ```
 
    The first command generates a single mutant, and its source path is relative to `.`,
    the default source root. We can see that the reported paths in `mutants.log`,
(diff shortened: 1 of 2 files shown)


docs/gambit/gambit.md
```diff
@@ -242,12 +242,11 @@ Here are some examples of using the `--sourceroot` option.
    ```
 
    This should output the following:
-   <!-- Code output: using `pre` to avoid the Copy To Clipboard feature -->
-   <pre>
+   ```
```
mdgeorge4153, line 245: I'm fine with this change, although IIRC this code is automatically generated; is this change in sync with changes in the original source?

docs/gambit/mutation-verifier.md
```diff
@@ -42,10 +42,9 @@ certoraMutate --prover_conf path/to/prover/prover.conf --mutation_conf path/to/m
 ```{note}
 You must run `certoraMutate` from the root of the Solidity project directory.
 The files `prover.conf` and `mutation.conf`
-can be in their own directories.
+can be in their own directories but always within the project directory.
```
mdgeorge4153, line 45:
```suggestion
can be in their own directories, but must always be within the project directory.
```

mdgeorge4153 (approved):
Looks good, as long as it is in sync with the auto-generation

docs/gambit/gambit.md
chandrakananandi: You are right that we have a PR for auto generated doc, but that PR is on hold for a bit. When that PR lands, it is expected to handle this correctly (i.e., change it to ``` when needed). Here, this was just a copy paste error.

PR #154, opened 2023-10-11
Initial cut at direct storage access docs

jtoman:
Jira ticket: TODO

### DIFF
--- a/docs/cvl/expr.md
+++ b/docs/cvl/expr.md
@@ -466,3 +466,46 @@ written and then later cleared by Solidity (via the `pop()` function or the `del
 clear operation the slot will definitely hold 0, but the Prover will not make any assumptions
 about the value of the uninitialized slot which means they can be considered different.
 ```
+
+(direct-storage-access)=
+Direct Storage Access
+---------------------
+
+The value of contract state variables can be directly accessed from CVL. These direct
+storage accesses are written using the state variable names and struct fields defined
+in the contract. For example, to access the state variable `uint x` defined in the `currentContract`,
+one can simply write `currentContract.x`. More complex structs can be accessed by chaining field selects
+and array/map dereference operations together. For example, if the current contract has the following
+type definitions and state variables:
+
+```
+struct Foo {
+   mapping (address => uint[]) bar;
+}
+Foo[3] myState;
+```
+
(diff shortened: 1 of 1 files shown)


docs/cvl/expr.md
```diff
@@ -466,3 +466,46 @@ written and then later cleared by Solidity (via the `pop()` function or the `del
...
+A contract identifier (or `currentContract`) *must* be included in the direct storage access. In other
+words, writing just `myState[0].bar[addr][0]` will not work, even if `myState` is declared in the current contract.
+```
+
+Currently only primitive values (e.g., `uint`, `bytes3`, `bool`, enums, and user defined value types) can be directly accessed.
+Attempting to access more complex types will yield a type checking error. For example, attempting to access
+an entire array with `currentContract.myState[0].bar[addr]` will fail.
+
+```{note}
+Although entire arrays cannot be accessed, the _length_ of the arrays can be accessed with `.length`, e.g., `currentContract.myState[0].bar[addr].length`.
```
shellygr, line 504: but not for static arrays

shellygr (commented):
are there more caveats we should mention here?

looping in @mdgeorge4153 too

docs/cvl/expr.md
```diff
@@ -466,3 +466,46 @@ written and then later cleared by Solidity (via the `pop()` function or the `del
...
+```
+
+```{warning}
+Direct storage access is an experimental feature, and relies on several internal static analyses which can sometimes fail.
```
mdgeorge4153, line 508: Not for this PR, but I think at some point it would be nice to provide some high-level documentation for the analyses that can fail, how you know they've failed, what goes wrong if they do, and maybe some suggestions (if possible) about how to work around them.

mdgeorge4153, line 508:
```suggestion
Direct storage access is an experimental feature, and relies on several internal program analyses which can sometimes fail.
```

mdgeorge4153: TIOLI, I'm not sure if our users know what static analysis means.

mdgeorge4153 (approved):
I have a few small nits, but overall very nicely written.  It might be worth coordinating with Sitvanit, who has been adding examples - both to create new examples for this feature and also to see if any existing examples should be updated to use this feature.

PR #157, opened 2023-10-17
[DOC-347] progress on scene parametric rules

mdgeorge4153:
Jira ticket: [DOC-347] https://certora.atlassian.net/browse/DOC-347)

### DIFF
--- a/docs/cvl/invariants.md
+++ b/docs/cvl/invariants.md
@@ -5,11 +5,13 @@ Invariants describe a property of the state of a contract that is always
 expected to hold.
 
 ```{caution}
-Even if an invariant is verified, it may still be possible to violate it.  This
-is a potential source of {term}`unsoundness <unsound>`.  See {ref}`invariant-assumptions`
-for details.
+Certain features of invariants are {term}`unsound`: the invariant can be
+verified by the Prover, but it may still be possible for the contract to
+violate it.  The possible sources of unsoundness are {ref}`preserved`,
+{ref}`invariant-filters`, and {ref}`invariant-revert`.  Invariant proofs are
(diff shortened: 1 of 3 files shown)


docs/cvl/invariants.md
```diff
@@ -163,13 +138,25 @@ block, if any), inside a set of curly braces (`{ ... }`).  Each preserved block
 consists of the keyword `preserved` followed by an optional method signature, 
 an optional `with` declaration, and finally the block of commands to execute.
 
+```{note}
+Although invariants are now checked on methods of all contracts on the scene,
+the method signature in a `preserved` block only allows specifying the method
+signature of the method, and not the receiver contract.  If multiple contracts
+on the scene implement methods with the same signature, you must use the same
+`preserved` block for all of them.
```
shellygr, line 146: Let's make a ticket to ensure we allow this flexibility in the future

docs/prover/cli/options.md
```diff
@@ -86,6 +92,75 @@ If we want to verify only `withdraw_succeeds`, we run
 If we want to verify both `withdraw_succeeds` and `withdraw_fails`, we run  
 `certoraRun Bank.sol --verify Bank:Bank.spec --rule withdraw_succeeds withdraw_fails`
 
+(--method)=
+### `--method <method_signature> ...`
```
shellygr, line 96: I think just one is supported

shellygr (approved):
Some comments but mostly good to go

docs/cvl/invariants.md
mdgeorge4153: CERT-4003

PR #160, opened 2023-10-19
Explain what the length of a dynamic array is

nafur:
This PR just adds a simple clarification about `.length` on dynamic arrays. Given that we also have `.length` for bytes, it is not immediately clear whether an array's length refers to the number of elements in that array, or its length in bytes.

This change reflects my current guess. Please confirm that it actually is correct!


mdgeorge4153 (approved):
Thank you!

PR #163, opened 2023-10-25
better strings

urikirsh:
Previously, both the Prover release notes and the verification report release notes appears as release notes, which is ugly and less readable. Also, updated the naming of the report log to have the same convention as the prover log.


mdgeorge4153 (approved):
lgtm

PR #164, opened 2023-10-27
[DOC-351] Document control flow splitting

alexandernutz:
Add a conceptual explanation for the (control flow) splitting feature. This understanding is helpful for mitigating timeouts using the corresponding collection of cli options.

Jira ticket: https://certora.atlassian.net/browse/DOC-351
Link to generated documentation: https://certora-certora-prover-documentation--164.com.readthedocs.build/en/164/docs/prover/techniques/index.html#control-flow-splitting


docs/prover/cli/options.md
```diff
@@ -793,29 +793,111 @@ Advanced options that control control flow splitting
...
+
+When looking for a SAT result and observing an [SMT-type timeout](timeouts-introduction).
+
+**Example**
+
+```sh
+certoraRun Bank.sol --verify Bank:bank.spec --prover_args '-dontStopAtFirstSplitTimeout true'
+```
 
 (-smt_initialSplitDepth)=
```
otakar-trunda, line 875: There's some interaction between `-smt_initialSplitDepth` and `-depth` that's worth mentioning here. If I remember correctly, `-depth` overrides `-smt_initialSplitDepth`, e.g. with `-depth 5` and `-smt_initialSplitDepth 10` it will still only split to depth 5.

otakar-trunda (approved):
I left two minor comments, otherwise it looks good.

docs/prover/cli/options.md
alexandernutz: I see, good point -- adding a sentence accordingly

docs/prover/techniques/index.md
```diff
@@ -1,31 +1,121 @@
...
+                worklist.add([split_1, current_depth + 1])
+                worklist.add([split_2, current_depth + 1])
+return UNSAT
+```
+
+Intuitively, the algorithm explores the tree of all possible recursive
+splittings along a fixed sequence of split points up to the maximum splitting
+depth. We call the splits at maximum splitting depth *split leafs*. The
+exploration stops in either of the following three cases:
+ - if one split was found that is SAT (reasoning: if one split is SAT, then the
```
mdgeorge4153, line 82: I think `SAT` and `UNSAT` require some mental gymnastics.  I think something like "if a counterexample is found for any split, then the process stops" would be clearer.

docs/prover/techniques/index.md
```diff
@@ -1,31 +1,121 @@
...
+Control flow splitting (or short "splitting") is one of the techniques that the
+Certora Prover employs to speed up solving. In the remainder of this section, we
+will give an overview of how the technique works. This background should be
+helpful when using the settings described [here](control-flow-splitting-options)
+to prevent Prover timeouts.
+
+Splitting is best illustrated using the {term}`control flow graph` (CFG) of a given
+CVL rule.
+
+A single splitting step proceeds as follows:
```
mdgeorge4153, line 29: I think there is a high-level idea that might get lost in the details.  Essentially whenever we find a branch (e.g. an `if` statement or loop guard), we sometimes look for counterexamples in each branch separately.  We could also include a small example, something like:

````
For example, consider a rule with an `if` statement:
```cvl
rule example {
  ...
  if (owner == spender) {
    assert balance_after == balance_before;
  } else {
    assert balance_after == balance_before + amount;
  }
}
```
To simplify the search for a counterexample, the Prover may internally split this single rule into two rules:
```cvl
rule example_split_1 {
  ...
  require owner == spender;
  assert balance_after == balance_before;
}

rule example_split_2 {
  ...
  require owner != spender;
  assert balance_after == balance_before + amount;
}
```

A counterexample for either of the split rules will also be a counterexample for the original rule, and any counterexample for the original rule must violate one of the two split rules, so this splitting doesn't change the meaning of the rule.  However, in some cases the split rule is easier for the Prover to reason about.

docs/prover/techniques/index.md
```diff
@@ -1,31 +1,121 @@
...
+ - Setting the [initial splitting depth](-smt_initialSplitDepth) to a value 
+   above 0 will make the Prover skip the checking and immediately enumerate all 
+   splits up to that depth.
 
 (storage-and-memory-analysis)=
-# Storage and Memory Analysis
+# Storage and Memory analysis
 
 The Certora Prover works on EVM bytecode as its input. To the bytecode, the
 address space of both Storage and Memory are flat number lines. That two
```
mdgeorge4153, line 108: I don't think Storage and Memory should be capitalized.

docs/prover/techniques/index.md
```diff
@@ -1,31 +1,121 @@
...
 ["non-aliasing"](https://en.wikipedia.org/wiki/Aliasing_(computing)) argument
-requires reasoning about multiplications, additions, and hash functions. Certora
-Prover models this reasoning correctly, but this naive low-level modeling can
-quickly overwhelm SMT solvers. In order to handle storage efficiently, Certora
-Prover analyses Storage (Memory) accesses in EVM code in order to understand the
-Storage (Memory) layout, thus making information like "an update to mapping `x`
-will never overwrite the scalar variable `y`" much more obvious to the SMT
-solvers. For scaling SMT solving to larger programs, these simplifications are
-essential.
+requires reasoning about multiplications, additions, and hash functions. The
```
mdgeorge4153, line 113: New paragraph

docs/prover/techniques/index.md
```diff
@@ -1,31 +1,121 @@
...
+Certora Prover models this reasoning correctly, but this naive low-level
```
mdgeorge4153, line 114: I have some additional edits for this section but I see this PR doesn't really modify this section so I will address them another time.

spelling_wordlist.txt
```diff
@@ -112,11 +113,13 @@ precompiled
 preprocess
 preprocessing
 prestate
+prover
```
mdgeorge4153, line 116: Prover should always be capitalized, which should not be flagged as a spelling error.

docs/prover/cli/options.md
```diff
@@ -793,29 +793,117 @@ Advanced options that control control flow splitting
...
+### `--prover_args '-depth <number>'`
+
+**What does it do?**
+
+Sets the maximum splitting depth.
+
+**When to use it?** 
+
+When the deepest splits are too heavy to solve, but not too high in number,
+increasing this will lead to smaller, but more split leafs, which run at the
```
mdgeorge4153, line 806:
```suggestion
increasing this will lead to smaller, but more split leaves, which run at the
```

docs/prover/techniques/index.md
```diff
@@ -1,31 +1,121 @@
...
+            else
+                [split_1, split_2] = split_single(cfg)
+                worklist.add([split_1, current_depth + 1])
+                worklist.add([split_2, current_depth + 1])
+return UNSAT
+```
+
+Intuitively, the algorithm explores the tree of all possible recursive
+splittings along a fixed sequence of split points up to the maximum splitting
+depth. We call the splits at maximum splitting depth *split leafs*. The
```
mdgeorge4153, line 80:
```suggestion
depth. We call the splits at maximum splitting depth *split leaves*. The
```

docs/prover/techniques/index.md
```diff
@@ -1,31 +1,121 @@
...
+exploration stops in either of the following three cases:
```
mdgeorge4153, line 81:
```suggestion
exploration stops in any of the following three cases:
```

docs/prover/cli/options.md
```diff
@@ -793,29 +793,117 @@ Advanced options that control control flow splitting
...
+certoraRun Bank.sol --verify Bank:bank.spec --prover_args '-depth 5'
+```
 
 (-mediumTimeout)=
 ### `--prover_args '-mediumTimeout <seconds>'`
 
-The "medium timeout" determines how much time is given to checking a split at
-not max-depth before we split again.
+The "medium timeout" determines how much time the SMT solver gets for checking a
+split that is not a split leaf.
```
mdgeorge4153, line 821: I would consider adding "split" and "split leaf" to the glossary with links to the new section.

docs/prover/cli/options.md
```diff
@@ -793,29 +793,117 @@ Advanced options that control control flow splitting
...
+Sets the time that non-leaf splits get before being split again. 
+
+**When to use it?** 
+
+When a little more time can close some splitting subtrees early, this can save a
+lot of time, since the subtree's size is exponential in the remaining depth. On
+the other hand, if something will be split further anyway, this can save the
+run time spent on intermediate "TIMEOUT" results. Use
+{ref}`-smt_initialSplitDepth` to eliminate that time investment altogether up to
+a given depth.
```
mdgeorge4153, line 834: This probably doesn't belong right here (but maybe in the main page on splitting), but this reminds me: Is there a way to know how much time was spent on different splits to help with this kind of tuning?

docs/prover/cli/options.md
```diff
@@ -793,29 +793,117 @@ Advanced options that control control flow splitting
...
+
+**Example**
+
+```sh
+certoraRun Bank.sol --verify Bank:bank.spec --prover_args '-mediumTimeout 20'
+```
+
+
+
+% The regular SMT timeout determines how much time is maximally spent on checking
```
mdgeorge4153, line 844: not sure why this is commented out

docs/prover/cli/options.md
```diff
@@ -793,29 +793,117 @@ Advanced options that control control flow splitting
...
 (-dontStopAtFirstSplitTimeout)=
 ### `--prover_args '-dontStopAtFirstSplitTimeout <true/false>'`
 
-We can tell the Certora Prover to not stop when the first split has had a
-maximum-depth timeout. Note that this is only useful for SAT results, since for
-an overall UNSAT results, all splits need to be UNSAT, while for a SAT result it
-is enough that one split is UNSAT.
+**What does it do?**
+
+We can tell the Certora Prover to continue even when the first split has had a
```
mdgeorge4153, line 858: as above, split should be a glossary link here.

docs/prover/cli/options.md
```diff
@@ -793,29 +793,117 @@ Advanced options that control control flow splitting
...
+maximum-depth timeout. Note that this is only useful for {term}`SAT result`s,
```
mdgeorge4153, line 859: As above, I think SAT / UNSAT is almost always more confusing for users than something like "finds a counterexample" / "successfully proves".

docs/prover/cli/options.md
```diff
@@ -793,29 +793,117 @@ Advanced options that control control flow splitting
 
 See [here](control-flow-splitting) for an explanation of control flow splitting.
```
mdgeorge4153, line 794: I would shorten "Advanced options that control control flow splitting" to "Control flow splitting options".  It also looks like the header levels are inconsistent; take a look at the generated table of contents at the top of the page.

mdgeorge4153 (changes requested):
Looking good!  There's a bit much low-level detail, but I think that's unavoidable for this information, and I think it is well scoped.  I have some editorial comments.

docs/prover/techniques/index.md
alexandernutz: The idea of capitalizing was to indicate that I'm referring to the terms as defined by EVM, not to the general concepts (in which case I wouldn't be sure what the difference between the two even is exactly) ... 

Maybe I make them glossary references instead of capitalizing (first occurrence at least .. make new glossary entries, too, if necessary)?.. 
I'll try that for now, open to comments.

docs/prover/techniques/index.md
alexandernutz: the problem is that it's not always a counterexample, sometimes it's a "witness example" ...
 -- so I chose SAT because it's the more "stable" concept here ...
I made a glossary entry explaining the background there.
 -- not needing the mental gymnastics is of course better 🤔 

I can try how it looks with "counterexample", and on first mention, or in a footnote, mention witness examples and SAT/UNSAT, pointing to the glossary ...

docs/prover/techniques/index.md
```diff
@@ -1,31 +1,121 @@
...
 
-In this chapter we describe techniques used by the Certora Prover whose understanding can be relevant for an expert-level usage of the Prover.
+In this chapter, we describe some of the techniques used inside the Certora
+Prover. While this knowledge is not essential for using the Prover, it can
+sometimes be helpful when the Prover does not behave as expected, for instance
+in case of a timeout.
 
 (control-flow-splitting)=
-# Control Flow Splitting
+# Control flow splitting
```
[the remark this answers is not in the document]
alexandernutz: ah.. thanks!

spelling_wordlist.txt
alexandernutz: yeah, we already had `Prover` .. removing `prover`

docs/prover/cli/options.md
alexandernutz: uh, I knew that at some point .. in high school .. changing

docs/prover/techniques/index.md
alexandernutz: (hm, the spell checker seems fine with it though .. changing anyway ..)

docs/prover/cli/options.md
alexandernutz: doing that

docs/prover/cli/options.md
alexandernutz: right, so I'm linking every first occurrence under each option ... (or other "entry point")

docs/prover/cli/options.md
alexandernutz: header: right, much better ...

hm, the toc looks okay-ish to me -- "Control flow splitting options" is on one level with the other groups ("Options that control the soldiity compiler", "Advanced Options", etc .. though whether Advanced options is a great category might be subject for discussion ..)

docs/prover/cli/options.md
alexandernutz: removed and added an extra sentence to the description of `-mediumTimeout` right above

docs/prover/techniques/index.md
alexandernutz: Right, making this into an intro section -- should be nice!

docs/prover/cli/options.md
alexandernutz: yeah, rewrote for counterexample / witness example ... clunky as well, but perhaps clearer ..

docs/prover/cli/options.md
alexandernutz: It is on my list for things to implement in the new "Live Stats" tab. -- once we have that, we should document it here.

docs/prover/techniques/index.md
alexandernutz: I'll use the terms "EVM storage" and "EVM memory" instead of the capitalized Storage and Memory ..

docs/prover/techniques/index.md
alexandernutz: hm, after revisiting this again, I feel like for this very technical section, SAT and UNSAT might be best, since they're just the most concise thing to put e.g. into the pseudo code ...
 -- there is a glossary entry for them, which also relates them to the concepts of counterexample and witness example.

I'll also add a sentence to the top of this section mentioning this relationship
 -- let me know if you think this won't do ...

docs/prover/cli/options.md
```diff
@@ -949,7 +953,7 @@ This option disables the storage splitting optimization.
 
 
 (--allow_solidity_calls_in_quantifiers)=
-### --allow_solidity_calls_in_quantifiers
+### `--allow_solidity_calls_in_quantifiers`
```
shellygr, line 956: an aside - if you're faster than me... can you fix to `--prover_args` instead of `--settings` in line 938 `--settings -smt_groundQuantifiers=false`? just noticed it's out of date...

docs/prover/cli/options.md
```diff
@@ -967,34 +971,113 @@ error on encountering contract method calls in quantified expression bodies.
...
+### `--prover_args '-depth <number>'`
+
+**What does it do?**
+
+Sets the maximum splitting depth.
+
+**When to use it?** 
+
+When the deepest {term}`split`s are too heavy to solve, but not too high in
+number, increasing this will lead to smaller, but more {term}`split leaves`, which run
```
shellygr, line 989:
```suggestion
number, increasing this will lead to smaller, but more numerous {term}`split leaves`, which run
```

docs/prover/cli/options.md
```diff
@@ -967,34 +971,113 @@ error on encountering contract method calls in quantified expression bodies.
...
+at the full SMT timeout (as set by {ref}`--smt_timeout`). Conversely, if run
+time is too high because there are too many splits, decreasing this number means
+that more time is spent on fewer, but bigger split leaves.
```
shellygr, line 992: 1. do we have a tip on when to increase or decrease? 
2. should we state the current default (10)?

docs/prover/cli/options.md
```diff
@@ -967,34 +971,113 @@ error on encountering contract method calls in quantified expression bodies.
...
+When a little more time can close some splitting subtrees early, this can save a
+lot of time, since the subtree's size is exponential in the remaining depth. On
+the other hand, if something will be split further anyway, this can save the
+run time spent on intermediate "TIMEOUT" results. Use
+{ref}`-smt_initialSplitDepth` to eliminate that time investment altogether up to
+a given depth.
+
+**Example**
+
+```sh
```
shellygr, line 1022: again, maybe mention default

shellygr (approved):
Great! nits

alexandernutz:
ugh, I guess I landed this rashly, will make an extra PR for your nits, Shelly

docs/prover/cli/options.md
alexandernutz: hm, the accent seems to be on master already; changed `--settings` (in separate branch, PR upcoming ..)

docs/prover/cli/options.md
alexandernutz: rright 💯

docs/prover/cli/options.md
alexandernutz: re 1: well .. this is supposed to be it ... but it's rather abstract, I admit .. I'll give it another shot
re 2: sure (increased maintenance effort / failure potential is a con, but I guess I'll add it and see how it goes ..)

alexandernutz: hm, actually, I'm still not sure what to write, as changing this should really be a reaction to where CVT is spending its time ... which currently can only be observed in statsdata.json ...
however, there are plans to condense this info to a live stat -- so I propose to update this when introducing these stats.

PR #169, opened 2023-10-31
fixed the build

mdgeorge4153:
Fixed a few spelling problems and other minor tweaks that broke the build


shellygr (approved):
Thanks for cleaning this up

PR #173, opened 2023-11-12
DOC-350: Add CERT-926 docs to the v5 changes doc set.

steve-certora:
### Description

This PR adds text to the v5 breaking changes markdown file in the mike/v5-breaking-changes branch and should be merged with that branch.  The PR contains a description of the changes, an example, a discussion of the motivation for the change, and a way to restore the previous behavior if needed.

Jira ticket: DOC-350
Link to generated documentation: [here](https://certora-certora-prover-documentation--173.com.readthedocs.build/en/173/docs/cvl/v5-changes.html)

### DIFF
--- a/docs/cvl/v5-changes.md
+++ b/docs/cvl/v5-changes.md
@@ -210,7 +210,7 @@ contract example {
 }
 ```
 
-```cvl
+```{code-block} cvl
 :emphasize-lines: 4
 
 rule for_all() {
--- a/docs/cvl/v5-changes.md
+++ b/docs/cvl/v5-changes.md
@@ -3,7 +3,7 @@ Certora CLI 5.0 Changes
 
 The release of `certora-cli` version 5.0 introduces a few small breaking
 changes for CVL.  These changes improve the coverage for parametric rules and
-invariants, remove support for Solidity function calls in quantified expressions,
+invariants, disallow Solidity function calls in quantified expressions,
 and simplify some rarely-used features.  This document explains those changes
 and how to work with them.
 
@@ -211,6 +211,8 @@ contract example {
 ```
 
 ```cvl
+:emphasize-lines: 4
+
 rule for_all() {
     // Using foo(i) in the quantified body will now cause the prover to
     // generate an error.
@@ -222,7 +224,7 @@ rule for_all() {
 In the example rule `for_all`, the prover will now generate an error similar
 to the following:
 
-```
+```txt
 Error in spec file (test2.spec:8:36): Contract function calls such as foo(i)
 are disallowed inside quantified formulas.
 ```
--- a/docs/cvl/v5-changes.md
+++ b/docs/cvl/v5-changes.md
@@ -224,7 +224,7 @@ rule for_all() {
 In the example rule `for_all`, the prover will now generate an error similar
 to the following:
 
-```txt
+```text
 Error in spec file (test2.spec:8:36): Contract function calls such as foo(i)
 are disallowed inside quantified formulas.
 ```
--- a/docs/cvl/v5-changes.md
+++ b/docs/cvl/v5-changes.md
@@ -198,7 +198,7 @@ Disallow Solidity function calls in quantified expressions
 ----------------------------------------------------------
 
 Starting with `certora-cli` version 5.0, the Prover no longer supports
-making Solidity contract `method` calls in quantified expression bodies by
+making Solidity contract method calls in quantified expression bodies by
 default. For example, given the simple contract below, you can no longer
 use the `method` `foo()` in a quantified expression body.
 
@@ -215,7 +215,7 @@ contract example {
 
 rule for_all() {
     // Using foo(i) in the quantified body will now cause the prover to
-    // generate an error.
+    // generate a type-checking error.
     require (forall uint256 i . i == foo(i));
     assert false, "Prover will generate an error before this line";
 }
--- a/docs/cvl/v5-changes.md
+++ b/docs/cvl/v5-changes.md
@@ -194,49 +194,68 @@ is the main contract being verified, then passing `--contract C` will cause
 method variables to be instantiated in the same way the would have in older
 versions.
 
-Disallow Solidity function calls in quantified expressions
-----------------------------------------------------------
+Disallow calls to contract functions in quantified expressions
+--------------------------------------------------------------
 
 Starting with `certora-cli` version 5.0, the Prover no longer supports
-making Solidity contract method calls in quantified expression bodies by
-default. For example, given the simple contract below, you can no longer
-use the `method` `foo()` in a quantified expression body.
+making contract method calls in quantified expression bodies by
+default.
+
+For example, given the simple contract below, you can no longer
+use the `method` `getI()` in a quantified expression body.
 
 ```{code-block} solidity
 contract example {
-   function foo(uint256 i) public pure returns (uint256) {
+
+   uint i; 
+    
+   function getI() public view returns (uint256) {
        return i;
-   } 
+   }
 }
 ```
 
 ```{code-block} cvl
 :emphasize-lines: 4
 
-rule for_all() {
-    // Using foo(i) in the quantified body will now cause the prover to
+rule there_exists {
+    // Using getI() in the quantified body will now cause the Prover to
     // generate a type-checking error.
-    require (forall uint256 i . i == foo(i));
+    require (exists uint256 i . i == getI());
     assert false, "Prover will generate an error before this line";
 }
 ```
 
-In the example rule `for_all`, the prover will now generate an error similar
+In the example rule `there_exists`, the Prover will now generate an error similar
 to the following:
 
 ```text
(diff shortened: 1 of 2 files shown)


docs/cvl/v5-changes.md
```diff
@@ -193,6 +194,52 @@ is the main contract being verified, then passing `--contract C` will cause
 method variables to be instantiated in the same way the would have in older
 versions.
 
+Disallow Solidity function calls in quantified expressions
+----------------------------------------------------------
+
+Starting with `certora-cli` version 5.0, the Prover no longer supports
+making Solidity contract `method` calls in quantified expression bodies by
```
shellygr, line 201: not sure we need to backtick `method` here

docs/cvl/v5-changes.md
```diff
@@ -193,6 +194,52 @@ is the main contract being verified, then passing `--contract C` will cause
...
+   } 
+}
+```
+
+```{code-block} cvl
+:emphasize-lines: 4
+
+rule for_all() {
+    // Using foo(i) in the quantified body will now cause the prover to
+    // generate an error.
```
shellygr, line 218: generate a type-checking error

shellygr (commented):
factually all is correct, just a few comments.
I'll let @mdgeorge4153 comment&approve on writing style specifics and any other comments Mike may have

docs/cvl/v5-changes.md
steve-certora: Changed.

docs/cvl/v5-changes.md
steve-certora: Changed.

docs/cvl/v5-changes.md
```diff
@@ -3,8 +3,9 @@ Certora CLI 5.0 Changes
 
 The release of `certora-cli` version 5.0 introduces a few small breaking
 changes for CVL.  These changes improve the coverage for parametric rules and
-invariants, and simplify some rarely-used features.  This document explains
-those changes and how to work with them.
+invariants, disallow Solidity function calls in quantified expressions,
+and simplify some rarely-used features.  This document explains those changes
+and how to work with them.
```
mdgeorge4153, line 8: I'm ok with this, although for this high-level summary I'm willing to consider the quantifier change to be included in "simplify some rarely-used features"

docs/cvl/v5-changes.md
```diff
@@ -193,6 +194,52 @@ is the main contract being verified, then passing `--contract C` will cause
 method variables to be instantiated in the same way the would have in older
 versions.
 
+Disallow Solidity function calls in quantified expressions
```
mdgeorge4153, line 197: I might use "calls to contract functions" instead of "Solidity function calls"

docs/cvl/v5-changes.md
```diff
@@ -193,6 +194,52 @@ is the main contract being verified, then passing `--contract C` will cause
...
+   function foo(uint256 i) public pure returns (uint256) {
+       return i;
+   } 
+}
+```
+
+```{code-block} cvl
+:emphasize-lines: 4
+
+rule for_all() {
```
mdgeorge4153, line 216: I may be in the minority, but I generally leave off the parentheses for rules:

```
rule for_all {
```

docs/cvl/v5-changes.md
```diff
@@ -193,6 +194,52 @@ is the main contract being verified, then passing `--contract C` will cause
...
+    // Using foo(i) in the quantified body will now cause the prover to
+    // generate a type-checking error.
+    require (forall uint256 i . i == foo(i));
+    assert false, "Prover will generate an error before this line";
+}
+```
+
+In the example rule `for_all`, the prover will now generate an error similar
```
mdgeorge4153, line 224:
```suggestion
In the example rule `for_all`, the Prover will now generate an error similar
```

mdgeorge4153: I'm surprised this passed the spell checker - somebody must have added lowercase-p prover to the dictionary.  I'll remove it

docs/cvl/v5-changes.md
```diff
@@ -193,6 +194,52 @@ is the main contract being verified, then passing `--contract C` will cause
...
 
+Disallow Solidity function calls in quantified expressions
+----------------------------------------------------------
+
+Starting with `certora-cli` version 5.0, the Prover no longer supports
+making Solidity contract method calls in quantified expression bodies by
+default. For example, given the simple contract below, you can no longer
+use the `method` `foo()` in a quantified expression body.
+
+```{code-block} solidity
```
mdgeorge4153, line 205: I don't feel strongly about this, but I would consider condensing this; something like "if `foo` is a contract method, then a quantified expression like `require (forall uint i . i == foo(i))` will no longer compile.

mdgeorge4153: Although on second though, you may want to keep the contract code so that you can refer to it below; I'll add a comment there.

docs/cvl/v5-changes.md
```diff
@@ -193,6 +194,52 @@ is the main contract being verified, then passing `--contract C` will cause
 method variables to be instantiated in the same way the would have in older
 versions.
 
+Disallow Solidity function calls in quantified expressions
+----------------------------------------------------------
+
+Starting with `certora-cli` version 5.0, the Prover no longer supports
+making Solidity contract method calls in quantified expression bodies by
+default. For example, given the simple contract below, you can no longer
```
mdgeorge4153, line 202: I would start a new paragraph for the example

docs/cvl/v5-changes.md
```diff
@@ -193,6 +194,52 @@ is the main contract being verified, then passing `--contract C` will cause
...
+to the following:
+
+```text
+Error in spec file (test2.spec:8:36): Contract function calls such as foo(i)
+are disallowed inside quantified formulas.
+```
+
+Using Solidity `method` calls in the quantified expressions introduced an
+element of instability to quantifier expressions that you can better
+overcome using ghosts, direct storage access, and other access patterns.
```
mdgeorge4153, line 234: There are two ideas in this paragraph.  The first is why we made the change, and the second is how to work around it.  I would probably either remove the first part or move it to the first paragraph in the section.  I would expand on the second point; the advice here isn't really actionable.  I would probably demonstrate how to rewrite the `foo` example above with both ghosts and direct storage access.

E.g. something like

"In most simple cases, you can replace the Solidity function call with either a {ref}`direct storage access <...>` or a {ref}`ghost <ghosts>`.  For example, the above function `foo` simply returns the storage variable `i`, so the `require` statement in the `for_all` rule could be replaced by `require (forall uint i . i == currentContract.i)`.  Alternatively, you could create a ghost variable that tracks `foo`:

```cvl
...
```

mdgeorge4153: I realize on second glance that this doesn't actually describe what `foo` does, and maybe this is harder to do for that example, so I might change the example a bit.

docs/cvl/v5-changes.md
```diff
@@ -193,6 +194,52 @@ is the main contract being verified, then passing `--contract C` will cause
...
+
+
+If you must use Solidity `method` calls in quantified expressions,
+you can still access the old behavior by specifying the
+`--allow_solidity_calls_in_quantifiers` argument to `certoraRun` on the
```
mdgeorge4153, line 239: For command line arguments, always use a `ref` and add a label to `docs/prover/cli/options.md`.  In this case:

```
specifying the {ref}`--allow_solidity_calls_in_quantifiers` argument ...
```

mdgeorge4153 (changes requested):
Looking good.  I have a few style comments and a request for more clarity in one section

docs/cvl/v5-changes.md
steve-certora: I do not have a preference, so if someone wants to weigh in, now is the time.

docs/cvl/v5-changes.md
steve-certora: Changed.

docs/cvl/v5-changes.md
steve-certora: Removed.

docs/cvl/v5-changes.md
steve-certora: Uppercased.

docs/cvl/v5-changes.md
steve-certora: Started.

docs/cvl/v5-changes.md
steve-certora: I reworked the example and removed the explanation for the change.

docs/cvl/v5-changes.md
steve-certora: I added the `ref` and an entry to `docs/prover/cli/options.md`.

steve-certora:
The current CI build is failing because of changes in the parent branch.

```
/home/docs/checkouts/readthedocs.org/user_builds/certora-certora-prover-documentation/checkouts/173/docs/cvl/v5-changes.md:188: WARNING: undefined label: --parametric_contracts
```

PR #177, opened 2023-12-05
[DOC-367] persistent ghosts explained with examples

shellygr:
Before requesting review:
 - Make sure CI is passing

Jira ticket: TODO

### DIFF
--- a/docs/cvl/ghosts.md
+++ b/docs/cvl/ghosts.md
@@ -6,17 +6,25 @@ Ghosts are a way of defining additional variables for use during verification.
 These variables are often used to communicate information between
 [rules](rules.md) and [hooks](hooks.md).
 
+Ghosts can be seen as an 'extension' to the state of the contracts under verification.
+This means that in case a call reverts, the ghost values will revert to their pre-state.
+Additionally, if an unresolved call is handled by a havoc, the ghost values will havoc as well.
+An exception to this rule are ghosts marked _persistent_.
+
 ```{contents}
 ```
 
 Syntax
 ------
 
-The syntax for ghost declarations is given by the following [EBNF grammar](syntax):
+The syntax for ghost and persistent ghost declarations is given by the following [EBNF grammar](syntax):
 
 ```
 ghost ::= "ghost" type id                             (";" | "{" axioms "}")
         | "ghost" id "(" cvl_types ")" "returns" type (";" | "{" axioms "}")
+
+persistent_ghost ::=  "persistent" "ghost" type id                             (";" | "{" axioms "}")
+                    | "persistent" "ghost" id "(" cvl_types ")" "returns" type (";" | "{" axioms "}")
 
 type ::= basic_type
        | "mapping" "(" cvl_type "=>" type ")"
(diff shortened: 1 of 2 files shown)


docs/cvl/ghosts.md
```diff
@@ -127,5 +135,97 @@ This documentation is currently incomplete.  See [ghosts](/docs/confluence/anato
...
+contract NotReentrant {
+    function transfer1Token(IERC20 a) external {
+        require (address(a) != address(this));
+        a.transfer(msg.sender, 1);
+    }
+}
+```
+
+If we do not apply any linking or dispatching for the call done on the target `a`, the call to `transfer` would havoc.
+In the lower-level view of the tool, the sequence of events that happens is as follows:
```
jtoman, line 179: "the sequence of events is as follows"

docs/cvl/ghosts.md
```diff
@@ -127,5 +135,97 @@ This documentation is currently incomplete.  See [ghosts](/docs/confluence/anato
...
+	function emptyRequire(uint a) external {
+		require(a != 0);
+	}
+}
+```
 
+It is expected that the method `userDefinedRequireMsg` and only it will satisfy the rule. 
+However, if we use regular ghosts, the ghost variable `saw_user_defined_revert_msg` will revert in case the input argument `a` is equal to 0, together with the contract itself. 
+This means the value of `saw_user_defined_revert_msg` will remain false, and thus the `satisfy` statement will fail.
+With regular ghosts, the rule would of course fail also for the two other methods, meaning that non-persistent ghosts are not suitable for distinguishing between different reverts.
```
jtoman, line 231: I've read this so many times, and I still have no idea what it's saying.

jtoman (changes requested):

docs/cvl/ghosts.md
```diff
@@ -6,18 +6,26 @@ Ghosts are a way of defining additional variables for use during verification.
 These variables are often used to communicate information between
 [rules](rules.md) and [hooks](hooks.md).
 
+Ghosts can be seen as an 'extension' to the state of the contracts under verification.
+This means that in case a call reverts, the ghost values will revert to their pre-state.
+Additionally, if an unresolved call is handled by a havoc, the ghost values will havoc as well.
```
nd-certora, line 11: Also, ghosts are regarded as part of the state and on  `at storage` statements are restored

docs/cvl/ghosts.md
```diff
@@ -127,5 +135,97 @@ This documentation is currently incomplete.  See [ghosts](/docs/confluence/anato
...
+    function transfer1Token(IERC20 a) external {
+        require (address(a) != address(this));
+        a.transfer(msg.sender, 1);
+    }
+}
+```
+
+If we do not apply any linking or dispatching for the call done on the target `a`, the call to `transfer` would havoc.
+In the lower-level view of the tool, the sequence of events that happens is as follows:
+1. A call to `a.transfer` which cannot be resolved and results in a havoc operation. Non-persistent ghosts are havoced, in particular `reentrancy_happened`.
```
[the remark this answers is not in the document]
shellygr: I actually want to show the correct version

docs/cvl/ghosts.md
```diff
@@ -127,5 +135,97 @@ This documentation is currently incomplete.  See [ghosts](/docs/confluence/anato
...
+2. A `CALL` hook executes, updating `reentrancy_happened` based on its havoced value, meaning it can turn to true.
+
+Therefore even if the addresses of `a` and `NotReentrant` are distinct, we could still falsely detect a reentrant call as `reentrancy_happened` was set to true due to non-determinism.
```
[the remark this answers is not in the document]
shellygr: reentrancy actually cannot occur. It does not matter because reentrancy_happened is in some state we track based on contracts, not tied to a particular contract. I added the wordier explanation, although for me it's harder to grasp

docs/cvl/ghosts.md
```diff
@@ -127,5 +135,97 @@ This documentation is currently incomplete.  See [ghosts](/docs/confluence/anato
...
+		require(a != 0, "a != 0");
+	}
+
+	function emptyRequire(uint a) external {
+		require(a != 0);
+	}
+}
+```
 
+It is expected that the method `userDefinedRequireMsg` and only it will satisfy the rule.
```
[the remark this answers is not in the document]
shellygr: "will satisfy". tried to reword

docs/cvl/ghosts.md
shellygr: tried to reword

PR #179, opened 2023-12-07
[DOC-370] Index page update

rolldavid:
Hide table of contents on index page but keep for the side nav


urikirsh:
This PR is no longer relevant: the new page `index.rst` already hides the `toctrees`.

PR #186, opened 2023-12-31
[DOC-373] document optimistic_summary_recursion and summary_recursion_limit

naftali-g:
[Generated "Options affecting summarization"](https://certora-certora-prover-documentation--186.com.readthedocs.build/en/186/docs/prover/cli/options.html#options-regarding-summarization)

### DIFF
--- a/docs/prover/cli/options.md
+++ b/docs/prover/cli/options.md
@@ -387,34 +387,64 @@ Options regarding summarization
 ### `--optimistic_summary_recursion`
 
 **What does it do?**
-In case there's a call to some Solidity function within a summary, we may end up with recursive calls to this summary. For example, if in the summary of `foo` we call the Solidity function `bar`, and `bar`'s Solidity code contains a call to `foo`, we'll summarize `foo` again, which will lead to another call to `bar` etc. In this case if this flag is set to `false` we may get an assertion failure with a message along the lines of
-```
+In case there's a call to some Solidity function within a summary, we may end up
+with recursive calls to this summary. For example, if in the summary of `foo` we
+call the Solidity function `bar`, and `bar`'s Solidity code contains a call to
+`foo`, we'll summarize `foo` again, which will lead to another call to `bar`
+etc. In this case if this flag is set to `false` we may get an assertion failure
+with a message along the lines of
+```text
 Recursion limit (...) for calls to ..., reached during compilation of summary ...
 ```
-Alternatively, such recursion could happen with {ref}`dispatcher` summaries - if both contract A and B have a function `foo()`, and `A.foo()` contains a call to `someContractAddress.foo()`, then if `foo()` is summarized with {ref}`dispatcher`, and `B.foo()` is unresolved, the dispatching may choose `A.foo()` as one of the possible targets and we end up with recursion. In this case if this flag is set to `false` we may get an assertion failure message along the lines of
-```
+Such recursion can also happen with {ref}`dispatcher summaries <dispatcher>` &mdash;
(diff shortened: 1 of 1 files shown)


docs/prover/cli/options.md
```diff
@@ -369,17 +369,53 @@ certoraRun Bank.sol --verify Bank:Bank.spec --optimistic_loop
...
+
+(--optimistic_summary_recursion)=
+### `--optimistic_summary_recursion`
+
+**What does it do?**
+In case there's a call to some Solidity function within a summary, we may end up with recursive calls to this summary. For example, if in the summary of `foo` we call the Solidity function `bar`, and `bar`'s Solidity code contains a call to `foo`, we'll summarize `foo` again, which will lead to another call to `bar` etc.
+The default behavior in this case is to assert if the recursion limit is reached (the limit is controlled by the {ref}`--summary_recursion_limit` flag). Setting this flag to `true` will instead assume that the limit is never reached.
+
+**When to use it**
+There's recursion due to summaries calling Solidity functions, and this leads to an assertion failure due to this. In this case one can either make the limit larger (via {ref}`--summary_recursion_limit`) or set this flag to `true`.
```
shellygr, line 394: can we make this even easier to users - and say which assert message they will see? I believe it's this:
`Recursion limit ... for calls to ..., reached during compilation of summary ...`

shellygr (commented):
shouldn't we add for `--contract_recursion_limit` or not yet because it's stil not in the python args?

naftali-g:
> shouldn't we add for `--contract_recursion_limit` or not yet because it's still not in the python args?

Yeah I don't think we should document non-existing flags :)

docs/prover/cli/options.md
```diff
@@ -369,17 +369,60 @@ certoraRun Bank.sol --verify Bank:Bank.spec --optimistic_loop
...
+**Example**
+
+```
+certoraRun Bank.sol --verify Bank:Bank.spec --optimistic_summary_recursion true
+```
+(--summary_recursion_limit)=
+### `--summary_recursion_limit`
+
+**What does it do?**
+In case there's a call to some Solidity function within a summary, we may end up with recursive calls to this summary. For example, if in the summary of `foo` we call the Solidity function `bar`, and `bar`'s Solidity code contains a call to `foo`, we'll summarize `foo` again, which will lead to another call to `bar` etc.
```
mdgeorge4153, line 412: I think instead of duplicating the text here, it might be better to just reference it.  E.g. ``Summaries can cause recursion (see {ref}`--optimistic_summary_recurion`).  This option sets the summary recursion level, which is the number of recursive calls that the Prover will consider.

If a counterexample causes a function to be called recursively more than the summary recursion limit, it will report an assertion failure (unless {ref}`--optimistic_summary_recursion` is set, in which case the counterexample will be ignored).

mdgeorge4153 (dismissed):
Looks good, I suggested a few changes.

mdgeorge4153:
It might also be worth adding a note about these options in the loop unrolling page (`docs/prover/approx/loops.md`), and possibly cross referencing the loop unrolling options (just to mention that they are parallel).

naftali-g:
> It might also be worth adding a note about these options in the loop unrolling page (`docs/prover/approx/loops.md`), and possibly cross referencing the loop unrolling options (just to mention that they are parallel).

I'm not sure what exactly to say about that :/ they seem parallel, but happen in completely different contexts (this isn't about recursion within contract code after all)

shellygr (approved):

PR #187, opened 2024-01-01
[DOC-372] Direct storage havoc

pickx:
[Jira ticket](https://certora.atlassian.net/browse/DOC-372)
[Link to generated documentation](https://certora-certora-prover-documentation--187.com.readthedocs.build/en/187/)


pickx:
from existing direct storage access documentation:

> For example, if the `myState` definition above appeared in
a contract called `Test` and the current CVL file included `using Test as t;` one could write `t.myState[0].bar[addr][0]`.

I'm guessing we're already aware of this, but from my testing, `t.myState[0].bar[addr][0]` doesn't actually compile right now. I think this has to do `myState` being not public, which seems to make compilation fail for nested direct storage accesses. 

(actually, I think you can't even make `myState` public, because `struct Foo` contains a dynamically-sized type? I don't know EVM well enough to understand the compile errors I'm seeing.)

this makes me wary of using this existing contract as a basis for examples, in case a user actually tries it.

docs/cvl/expr.md
```diff
@@ -515,3 +517,16 @@ Direct storage access is an experimental feature, and relies on several internal
 If these internal static analyses fail, any rules that use direct storage access will fail during processing. If this
 occurs, check the "Global Problems" view of the web report and contact Certora for assistance.
 ```
+
+### Direct storage havoc
+
+The same direct storage syntax can also be used in `havoc` statements. With the previously-mentioned `Example` contract and `using Example as ex`, you can write `havoc ex.luckyNumber` or `havoc addresses[10]` or even `havoc addresses.length`.
+
+While you may use a `havoc assuming` statement, unlike [ghosts](ghosts), you cannot directly refer to the havoced storage path in the `assuming`` expression using the `@old` and `@new` syntax. This generally means `assuming` expressions are not as useful with direct storage access, so consider using and unconditional `havoc` statements instead of `havoc assuming`.
```
naftali-g, line 525:
```suggestion
While you may use a `havoc assuming` statement, unlike [ghosts](ghosts), you cannot directly refer to the havoced storage path in the `assuming` expression using the `@old` and `@new` syntax. This generally means `assuming` expressions are not as useful with direct storage access, so consider using and unconditional `havoc` statements instead of `havoc assuming`.
```

naftali-g (changes requested):
Did you address @shellygr 's comment?

pickx:
> Did you address @shellygr 's comment?

other way around, it's a comment for Shelly (or anyone else who wants to take a look, I just tagged the first person that came to mind)

naftali-g (approved):

pickx:
after speaking with @jtoman - what seems to be happening here is this:
the storage data we use to match the storage path comes from the storage splitter (see `toStorageVar` in `StorageAccessCompiler.kt`). if a storage slot is unused, the splitter will miss it. 

for example, given
```
contract Example {
   address[] addresses;
}
```
with `using Test as t;`, attempting to access `t.addresses[0]` will crash with an unhelpful message.
if we force the splitter to see `addresses`, such as by making it `public` or by adding a function to the contract which interacts with this variable, the access will resolve as expected.

I'm going to mention this in the documentation. since there are no real blockers here, I'll then go ahead and merge.

PR #203, opened 2024-01-26
DOC-384: We can only hook once

nafur:
Add a paragraph that explains that we can not hook anything more than once.

Jira ticket: https://certora.atlassian.net/browse/DOC-384
Link to generated documentation: https://certora-certora-prover-documentation--203.com.readthedocs.build/en/203/docs/cvl/hooks.html


docs/cvl/hooks.md
```diff
@@ -46,6 +46,13 @@ See {doc}`statements` for information about the `statement` production; see
 {doc}`types` for the `evm_type` production; see {doc}`basics` for the `number`
 production.
 
+It is generally prohibited to hook more than once on any specific opcode or
+storage location to avoid ambiguities on the order of hook applications.
```
shellygr, line 50: This is very academic, so maybe add some simple sentences like: "The typechecker will emit a suitable error if multiple hooks on the same pattern are defined"

shellygr (approved):

docs/cvl/hooks.md
nafur: I rewrote the text to be simpler (not sure how well that worked out...)
I also realized that my claim that access path resolution is injective (and thus access paths are different if they are syntactically different) is not true.

docs/cvl/hooks.md
```diff
@@ -46,6 +46,18 @@ See {doc}`statements` for information about the `statement` production; see
...
+It is prohibited to have multiple hooks with the same hook pattern.
+Two hooks have the same hook pattern if both are `Sstore` hooks with the same
+access path, both are `Sload` hooks with the same access path, or both are
+opcode hooks with the same opcode.
+Doing so will result in an error like this:
+`The declared hook pattern <second hook> duplicates the hook pattern <first hook> at <spec file>. A hook pattern may only be used once.`
+Note that two access paths are considered to be "the same" if they resolve to
+the same storage address. Syntactically different access paths can alias when
+accessing a member by name (`contract.member`), by slot (`contract.(slot n)`),
+or by offset (`contract.(offset n)`).
```
shellygr, line 58: It's probably `slotPrefix.(offset n)` as an offset cannot be accessed from a contract directly, but with respect to some slot

shellygr (approved):
nit on offsets

PR #205, opened 2024-01-26
Adding docs for keccak256 and ecrecover in CVL

shellygr:
generated docs https://certora-certora-prover-documentation--205.com.readthedocs.build/en/205/


docs/cvl/expr.md
```diff
@@ -550,3 +550,116 @@ As with direct storage access in general, direct storage havoc is experimental a
...
+## Built-in Functions
+
+### Hashing
+
+Many contracts employ hashing in order to create `mapping` keys from more complex data structures.
+A native hashing function in Solidity is `keccak256`.
+We expose `keccak256` in CVL, in order to be able to create such mapping keys directly in CVL.
+The alternative to using the CVL built-in for `keccak256` is to use pre-existing
+functionality from the contract, or creating a dedicated harness function,
+which is usually undesirable from both usability and Prover performance perspectives.
```
jtoman, line 563: this seems weirdly specific. I don't know we need to belabor why we add it. Simply describe the primitive, and **then** give an example of how it can be used. As is, it sounds like `keccak256` can or should only be used for mapping keys.

docs/cvl/expr.md
```diff
@@ -550,3 +550,116 @@ As with direct storage access in general, direct storage havoc is experimental a
...
+  return keccak256(s.signature);
+}
+
+function hashingScheme3CVL(HashingExample.SignedMessage s) returns bytes32 {
+  return keccak256(s.sender, s.nonce, keccak256(s.signature));
+}
+```
+
+The scheme implemented in `hashingScheme4` is not supported at the moment, as it combines a `bytes` type with primitives.
+In general, the `keccak256` built-in function supports two kinds of inputs:
```
jtoman, line 614: I'm not sure what "in general" means here, it suggests that there are some exceptions to this rule, which there are not.

docs/cvl/expr.md
```diff
@@ -550,3 +550,116 @@ As with direct storage access in general, direct storage havoc is experimental a
...
+  // uniqueness of signature
+  require (forall uint8 v. forall bytes32 r. forall bytes32 s. forall bytes32 h1. forall bytes32 h2.
+    h1 != h2 => ecrecover(h1, v, r, s) != 0 => ecrecover(h2, v, r, s) == 0);
+  // dependency on r and s
+  require (forall bytes32 h. forall uint8 v. forall bytes32 s. forall bytes32 r1. forall bytes32 r2.
+    r1 != r2 => ecrecover(h, v, r1, s) != 0 => ecrecover(h, v, r2, s) == 0);
+  require (forall bytes32 h. forall uint8 v. forall bytes32 r. forall bytes32 s1. forall bytes32 s2.
+    s1 != s2 => ecrecover(h, v, r, s1) != 0 => ecrecover(h, v, r, s2) == 0);
+}
+```
```
jtoman, line 647: reading this doc, I would wonder why this would not just be the default behavior.

docs/cvl/expr.md
```diff
@@ -550,3 +550,116 @@ As with direct storage access in general, direct storage havoc is experimental a
...
+### ECRecover
+
+The `ecrecover` function in Solidity is helpful in recovering the signer's address from a signed message.
+It exists in very similar form in CVL and receives exactly the same parameter types as its Solidity counterpart.
+
+```{note}
+`ecrecover` is ***supported*** in quantified expressions.
+```
+
+In the Prover, `ecrecover` is implemented as an {ref}`uninterpreted functions <uninterp-functions>` uninterpreted function.
```
jtoman, line 631: does this implementation detail matter to a user of the documentation? I would phrase it like:

```
The Prover's model of `ecrecover` does not actually implement the elliptical curve recovery algorithm, and is instead implemented using an `uninterpreted function`. Like all uninterpreted functions, axioms [have a link here defining what an axiom is? do we have a glossary?] can be added to make the behavior of CVL's `ecrecover` more faithfully model the actual key recovery algorithm.
```

Also, this documentation has `uninterpreted function` appear twice in a row, once as a link, and once as plain text.

jtoman (changes requested):

docs/cvl/expr.md
shellygr: We're concerned about performance in case grounding is disabled

docs/cvl/expr.md
shellygr: right, but in a way it's not exactly the same primitive as in solidity the way ecrecover is virtually the same

docs/cvl/expr.md
shellygr: great, then I read the code correctly.

docs/cvl/expr.md
```diff
@@ -550,3 +550,112 @@ As with direct storage access in general, direct storage havoc is experimental a
...
+### ECRecover
+
+The `ecrecover` function in Solidity is helpful in recovering the signer's address from a signed message.
+It exists in very similar form in CVL and receives exactly the same parameter types as its Solidity counterpart.
+
+```{note}
+`ecrecover` is ***supported*** in quantified expressions.
+```
+
+The Prover's model of `ecrecover` does not actually implement the elliptical curve recovery algorithm, and is instead implemented using an {ref}`uninterpreted functions <uninterp-functions>`. Like all uninterpreted functions, {ref}`axioms <glossary>` can be added to make the behavior of CVL's `ecrecover` more faithfully model the actual key recovery algorithm.
```
jtoman, line 627: `uninterpreted functions` -> `uninterpreted function` (singular)

jtoman (approved):
fine modulo one nit

docs/cvl/expr.md
shellygr: fixed

PR #217, opened 2024-02-10
CERT-4884 updated docs (pending v7 release)

shellygr:


docs/prover/cli/options.md
```diff
@@ -688,6 +688,54 @@ struct TokenPair {
 We have two contracts `BankToken.sol` and `LoanToken.sol`. We want `tokenA` of the `tokenPair` to be `BankToken`, and `tokenB` to be `LoanToken`. Addresses take up only one slot. We assume `tokenPair` is the first field of Bank (so it starts at slot zero). To do that, we use:
 `certoraRun Bank.sol BankToken.sol LoanToken.sol --verify Bank:Bank.spec --structLink Bank:0=BankToken Bank:1=LoanToken`
 
+(--contract_recursion_limit)=
+### `--contract_recursion_limit`
+
+**What does it do?**
+Contract inlining can cause recursion (see {ref}`--optimistic_contract_recursion`). This
+option sets the contract recursion level, which is the number of recursive calls
+that the Prover will consider when inlining contracts linked using e.g. `--link` or `--struct_link`.
```
aehyvari, line 697:
```suggestion
that the Prover will consider when inlining contracts linked using, e.g., `--link` or `--struct_link`.
```

docs/prover/cli/options.md
```diff
@@ -688,6 +688,54 @@ struct TokenPair {
...
+If a counterexample causes a function to be called recursively more than the
+contract recursion limit, it will report an assertion failure (unless
+{ref}`--optimistic_contract_recursion` is set, in which case the counterexample
+will be ignored).
+The default value is zero (i.e. no recursion is allowed).
+
+**When to use it**
+Use this option when there is recursion due to linking calling Solidity
+functions, and this leads to an assertion failure. In this case one can either
+make the limit larger or set (via {ref}`--optimistic_contract_recursion`) flag
```
aehyvari, line 708:
```suggestion
make the limit larger or set {ref}`--optimistic_contract_recursion` flag
```

docs/prover/cli/options.md
```diff
@@ -688,6 +688,54 @@ struct TokenPair {
...
+**What does it do?**
+Contract inlining can cause recursion (see {ref}`--optimistic_contract_recursion`). This
+option sets the contract recursion level, which is the number of recursive calls
+that the Prover will consider when inlining contracts linked using e.g. `--link` or `--struct_link`.
+
+If a counterexample causes a function to be called recursively more than the
+contract recursion limit, it will report an assertion failure (unless
+{ref}`--optimistic_contract_recursion` is set, in which case the counterexample
+will be ignored).
+The default value is zero (i.e. no recursion is allowed).
```
aehyvari, line 703:
```suggestion
The default value is zero (i.e., no recursion is allowed).
```

docs/prover/cli/options.md
aehyvari: What if the recursion happens without linking, for example if a function called in main contract (`currentContract`) is already recursive?  Maybe that's already considered linking, but it's not very clear.

docs/prover/cli/options.md
```diff
@@ -688,6 +688,54 @@ struct TokenPair {
...
+(--contract_recursion_limit)=
+### `--contract_recursion_limit`
+
+**What does it do?**
+Contract inlining can cause recursion (see {ref}`--optimistic_contract_recursion`). This
+option sets the contract recursion level, which is the number of recursive calls
+that the Prover will consider when inlining contracts linked using e.g. `--link` or `--struct_link`.
+
+If a counterexample causes a function to be called recursively more than the
+contract recursion limit, it will report an assertion failure (unless
```
aehyvari, line 700:
```suggestion
If the prover finds an execution in which a function is called recursively more
than the contract recursion limit, the prover will report an assertion failure (unless
```

docs/prover/cli/options.md
```diff
@@ -688,6 +688,54 @@ struct TokenPair {
...
+{ref}`--optimistic_contract_recursion` is set, in which case the counterexample
+will be ignored).
+The default value is zero (i.e. no recursion is allowed).
+
+**When to use it**
+Use this option when there is recursion due to linking calling Solidity
+functions, and this leads to an assertion failure. In this case one can either
```
aehyvari, line 707: I'm not sure I understand the "calling Solidity functions" attribute.  Maybe it should be

```suggestion
Use this option when after linking the resulting program calls recursive solidity
functions, and this leads to an assertion failure. In this case one can either
```

docs/prover/cli/options.md
```diff
@@ -688,6 +688,54 @@ struct TokenPair {
...
+**What does it do?**
+Contract linking can cause recursion (see also {ref}`--contract_recursion_limit`).
+This option sets the Prover to optimistically assume that recursion cannot go 
+beyond what is defined by {ref}`--contract_recursion_limit`, 
+but only if {ref}`--contract_recursion_limit` is set to a number higher than 0.
+
+**When to use it?**
+1. When the recursion due to contract linking is unbounded.
+2. When we are interested only in a limited recursion depth due to contract linking.
+
```
aehyvari, line 733:
```suggestion

{caution}
Note that this flag could be another cause for unsoundness - even if such recursion
_could_ actually happen in the deployed contract, this code-path won't be verified.
```

docs/prover/cli/options.md
```diff
@@ -688,6 +688,54 @@ struct TokenPair {
...
+### `--contract_recursion_limit`
+
+**What does it do?**
+Contract inlining can cause recursion (see {ref}`--optimistic_contract_recursion`). This
+option sets the contract recursion level, which is the number of recursive calls
+that the Prover will consider when inlining contracts linked using e.g. `--link` or `--struct_link`.
+
+If a counterexample causes a function to be called recursively more than the
+contract recursion limit, it will report an assertion failure (unless
+{ref}`--optimistic_contract_recursion` is set, in which case the counterexample
```
aehyvari, line 701:
```suggestion
{ref}`--optimistic_contract_recursion` is set, in which case the execution
```

docs/prover/cli/options.md
aehyvari: I suppose the assertion failure might be a real one, for example, reachable within the current contract recursion limit.  The message probably tells when this is the case (like it does for summary recursion, right?).

aehyvari (commented):
It could be good to cross-reference `--optimistic_summary_recursion` and `--summary_recursion_limit`.

docs/prover/cli/options.md
shellygr: In what sense recursive? Internal calls? I can clarify that's not the kind of recursion this option deals with, it deals with recursion from external calls.

nd-certora:
is this PR supposed to fix also  optimisticFallback ? 
https://docs.certora.com/en/latest/docs/prover/cli/options.html#prover-args-optimisticfallback-true

shellygr:
@nd-certora was sure it was added somewhere but can't see it. Re-applied

PR #218, opened 2024-02-10
Behavior of memory to storage copies

shellygr:


urikirsh:
Going over the stale open doc PRs. @shellygr  is this still relevant? If so, I'll review it

shellygr:
Should be still true

docs/user-guide/gaps.md
```diff
@@ -180,4 +180,138 @@ is sufficient both by running the basic sanity rule,
...
+The push to `myArray` generates the two aforementioned loops. 
+If we wish to analyze this code with the Prover, there are two questions to be answered:
+
+1. Do we need to set a value for `--loop_iter` which is bigger than 1?
+
+Yes - the `data` local variable is put into a `struct ScheduledExecution` that is
+put into an array in storage. 
+This is done by the compiler using a 'copy-loop'.
+The selector component `this.testPush.selector` requires one iteration.
+the string `aa` requires three iterations: according to the [ABI specification](https://docs.soliditylang.org/en/v0.8.24/abi-spec.html),
```
urikirsh, line 226:
```suggestion
The string `aa` requires three iterations: according to the [ABI specification](https://docs.soliditylang.org/en/v0.8.24/abi-spec.html),
```

docs/user-guide/gaps.md
```diff
@@ -180,4 +180,138 @@ is sufficient both by running the basic sanity rule,
...
+
+// Violates sanity:
+certoraRun MemoryToStorage.sol --verify MemoryToStorage:sanity.spec --loop_iter 2
+```
+
+2. Do we need to set `--optimistic_loop`?
+
+Yes - when we copy `data` from memory to storage, the Solidity compiler
+also generates code that nullifies the previous data. 
+As we do not know (and probably not wishing to constrain) the size of the previous data,
```
urikirsh, line 250:
```suggestion
As we do not know (and probably do not wish to constrain) the size of the previous data,
```

docs/user-guide/gaps.md
```diff
@@ -180,4 +180,138 @@ is sufficient both by running the basic sanity rule,
...
+    bytes memory data = abi.encodeWithSelector(
+      this.testPush.selector,
+      "aa"
+    );
+    myArray[0] = ScheduledExecution(where, executed, data);
+  }
+}
+```
+
+The only difference in this new functionality is that we assume that the previous data has size of 225 bytes.
```
urikirsh, line 302:
```suggestion
The only difference in this new functionality is that we assume that the previous data has a size of 225 bytes.
```

docs/user-guide/gaps.md
```diff
@@ -180,4 +180,138 @@ is sufficient both by running the basic sanity rule,
...
+    );
+    myArray.push(ScheduledExecution(where, executed, data));
+  }
+}
+```
+
+The push to `myArray` generates the two aforementioned loops. 
+If we wish to analyze this code with the Prover, there are two questions to be answered:
+
+1. Do we need to set a value for `--loop_iter` which is bigger than 1?
```
urikirsh, line 220: Isn't the prover smart enough to unroll the loop enough times for this case? The string is static

urikirsh (commented):
Small fixes and a question

PR #221, opened 2024-02-16
Update timeout prevention docs for live stats [DOC-402]

alexandernutz:
The new live stats panel contains links into parts of the timeout doc. In particular the links are from the rule-level statistics, i.e., path count, nonlinear op count, and memory complexity, into the respective “Dealing with …” section of the doc. Those sections should work well as entry points to the doc, coming from the report.

Jira ticket: https://certora.atlassian.net/browse/DOC-402
Link to generated documentation: https://certora-certora-prover-documentation--221.com.readthedocs.build/en/221/docs/user-guide/timeouts/timeouts-main.html#dealing-with-different-kinds-of-complexity


docs/user-guide/timeouts/timeouts-main.md
```diff
@@ -215,6 +215,10 @@ Some of the information in these references is out of date.
 
 ## Dealing with different kinds of complexity
 
+% screenshots in this subsection are taken from this run:
+% https://vaas-stg.certora.com/output/80942/9101c7e51a27456eb51bd9d088949c92?anonymousKey=25cca030b7594b795d994e937b5a027812d9406d
```
nd-certora, line 219: better to give link to prover.certora.com when it's ready

docs/user-guide/timeouts/timeouts-main.md
```diff
@@ -227,6 +231,24 @@ worth considering no matter which statistic is showing high severity.
 (high-path-count)=
 ### Dealing with a high path count
 
+The number of control flow paths is a major indication of how difficult a rule
```
nd-certora, line 234: we assume there is knowledge of what we call the control flow path. Most programmers think if it differently - without unrolling and without inlining, so our number look huge to them. Not sure however how easy is it to example, maybe we should have a webinar and then link to it

docs/user-guide/timeouts/timeouts-main.md
```diff
@@ -227,6 +231,24 @@ worth considering no matter which statistic is showing high severity.
...
+reports. There, the path count is given both globally, and per call.
+
+```{figure} path-count-stats.png
+:name: path-count-stats
+Global and per-call path counts are displayed in the Live Statistics panel for 
+each rule.
+```
+
+In order to reduce the path count of a rule, modularization techniques
+(typically: method summarization) can be applied. However, altering the
```
nd-certora, line 248: another source of increase of path count is the setup and specifically dispatcher(true) to many alternatives. Can we add a comment on that?

docs/user-guide/timeouts/timeouts-main.md
```diff
@@ -280,28 +302,80 @@ certoraRun ... --prover_args '-dontStopAtFirstSplitTimeout true -depth 15 -mediu
 ### Dealing with nonlinear arithmetic
 
 Nonlinear integer arithmetic is often the hardest part of the formulas that
-Certora Prover is solving.
+Certora Prover is solving. 
+
+The Certora Prover displays the absolute number of nonlinear operations, as well
+as their number per external call, in the Live Statistics panel. In the per-call
+display, there is a warning-sign next to the call when there is a non-trivial
```
nd-certora, line 309: this number includes nonlinear operations in calls from this function, right? both internal and externall

docs/user-guide/timeouts/timeouts-main.md
```diff
@@ -280,28 +302,80 @@ certoraRun ... --prover_args '-dontStopAtFirstSplitTimeout true -depth 15 -mediu
...
 
 (high-memory-complexity)=
 ### Dealing with high memory (or storage) complexity
 
-In this subsection we consider common culprits for high memory complexity.
+The memory complexity of each rule or parametric rule is displayed in the Live
+Statistics panel in the Certora Prover reports. Memory complexity is measured by
+two values *graph size*, and *longest path*. Graph size indicates how often
+memory is updated in the rule overall.  Longest path indicates how many updates
+any given internal memory variable (e.g. the storage of a given contract, or a
```
nd-certora, line 365: I'm totally confused now. so what's the graph and the path. I was under the impression that it is dependencies between updates, i.e, def-use chains. Is it just number of operations?

docs/user-guide/timeouts/timeouts-main.md
alexandernutz: this is just a comment - I don't want to publish this link

the more interesting overarching question: is it ok to show screenshots from customers? so far, most were taken from some element/delvtech run (the ones calling AaveHyperdrive stuff) -- the run in this link is an outlier ... (forgot now which project, but it was more recent ..

docs/user-guide/timeouts/timeouts-main.md
alexandernutz: ugh.. maybe we should drop the mem complexity heuristic altogether ... 
or have just one number 🤔 
or maybe explain less?

I'll sleep on it ..

docs/user-guide/timeouts/timeouts-main.md
alexandernutz: yeah, everything below this function is counted
so as if everything was inlined (should state this clearly), i.e., the numbers will always descend when going deeper into the tree
 --> will try to clarify

docs/user-guide/timeouts/timeouts-main.md
alexandernutz: happily, always feels a bit piecewise, when I'm listing such things ..

docs/user-guide/timeouts/timeouts-main.md
shellygr: actually it's useful. I understand it like this, feel free to rephrase or use:

> The Certora Prover performs decompilation of bytecode converting a stack-based machine to a register-based machine, that includes key-to-value mappings for EVM objects such as memory and storage. In some cases, it's able to simplify the mappings to use simpler register variables. 
> 
> Under this model, the "Graph size" is a measure of how many times we store into a key-value mapping such as memory, storage, or a ghost function.
> The "Path" is a measure of the longest chain of writes performed on one of the mappings.
> In both cases, a smaller number indicates a less difficult problem for the Prover to solve.

docs/user-guide/timeouts/timeouts-main.md
alexandernutz: rewrote things a little, added a sub-section with some explanations for the path count stat
 - mention that the CFG considered is the one with everything inlined
 - added a little picture with the diamond shapes, illustrating the exponential growth 
 - mention dispatcher summaries as an source of "diamonds"

docs/user-guide/timeouts/timeouts-main.md
alexandernutz: done (see above)

docs/user-guide/timeouts/timeouts-main.md
alexandernutz: used Shelly's text (hope I didn't butcher it too much ... or only in a good way :-) ) -- keeping the stat as is for now

nd-certora (approved):

PR #224, opened 2024-02-27
document the change in default summarization policy for external wildcard summaries

shellygr:
https://certora-certora-prover-documentation--224.com.readthedocs.build/en/224/


docs/cvl/methods.md
```diff
@@ -346,8 +346,18 @@ determine whether the call should be replaced by an approximation.
 To determine whether a function call is replaced by an approximation, the
 Prover considers the context in which the function is called in addition to the
```
alexandernutz, line 347: writing "together with" instead of "in addition to" would have helped me a good bit, I think 
 --> it's more like that the context (which is the "whether it's resolved status" of the function, or something like this, right??.. EDIT: ah, no .. it's not ...) is a parameter to the policy or so. The "in addition to" makes it feel like they're kind of independent factors. If that makes sense ..

... but I think the paragraph needs something extra anyway -- see below ..

docs/cvl/methods.md
```diff
@@ -346,8 +346,18 @@ determine whether the call should be replaced by an approximation.
...
+specified method, while `UNRESOLVED` applies only to methods that cannot be fully
+resolved (i.e., both target contract and method identifier are known).
+For internal summaries, the default is `ALL`, as all internal functions
+are always resolvable; thus `UNRESOLVED` is impossible and will yield an error. 
+Similarly, for external summaries with contract-specific entries, 
+the default policy is `ALL`. 
+Conversely, for any external summary on wildcard contracts, the default 
+policy is `UNRESOLVED`. One may apply the `ALL` policy to make the summary apply
+on all instances of the wildcard method, even on target contracts for which
+it was resolved, e.g. by {ref}`linking <--link>`.
```
alexandernutz, line 360: I think this whole paragraph is quite confusing.

I think the confusion has to do with the "context" and "resolved" parts not really being explained. (also couldn't find it further up the doc)

So, instead of talking about "context in which the function is called" abstractly, those contexts should be named.

If things don't fit one paragraph, it's ok to forward-reference like that in an overview-paragraph, but then a paragraph explaining "context" and "resolved" business should follow. Then I think this new paragraph will make more sense.

alexandernutz (changes requested):
I think some things need to be expanded / explained a bit for this new paragraph to be easily understood ...
(I can make a full-text proposal if it helps, though I'm not that firm with function resolution, so might need a little explanation myself ..)

docs/cvl/methods.md
shellygr: added explanation on contexts, on attributes defining a call site, and what it means to resolve a call in a particular call site

docs/cvl/methods.md
```diff
@@ -338,16 +338,56 @@ There are several kinds of summaries available:
 (delete-summary)=
 ### Summary application
 
-To decide whether to summarize a given internal or external function call, the
+To decide whether to summarize a given function call at a given call site, the
 Prover first determines whether it matches any of the declarations in the
-methods block, and then uses the declaration and the calling context to
+methods block, and then uses the declaration and the _calling context_ to
```
alexandernutz, line 343:
```suggestion
methods block, and afterward uses the declaration and the _calling context_ to
```
(gut feeling)

alexandernutz: the italics on calling context denote it to be a technical term, but that exact term does not appear again -- if "calling context" is indeed identical with "(3) The context in which it is called", then ideally, the term would be repeated syntactically
(feel free to skip if you think this is overly pedantic)
 --- i'll make a suggestion there ..

docs/cvl/methods.md
```diff
@@ -338,16 +338,56 @@ There are several kinds of summaries available:
...
+function callIt(address it) external {
+  IERC20(it).transfer(...); // cast `it` to an IERC20 contract and call the `transfer` method
+}
+```
+Similarly, the method signature may also be unresolvable:
+```solidity
+function callIt(bytes memory data) external {
+  address(this).call(data);
+}
+```
```
alexandernutz, line 373: luv to example 😍

docs/cvl/methods.md
alexandernutz: I think it's much clearer now! -- hopefully for others as well ..

alexandernutz (approved):
great! mini-nits

PR #246, opened 2024-04-08
DOC-415 | Add CI Configuration Guide

ozcodes:
Adding a small guide for customers on how to configure their ci to use certora cli in github actions.


docs/user-guide/ci.rst
```diff
@@ -0,0 +1,62 @@
...
+         run: |
+            wget https://github.com/ethereum/solidity/releases/download/v0.8.23/solc-static-linux
+            chmod +x solc-static-linux
+            sudo mv solc-static-linux /usr/local/bin/solc8.23
+
+         - name: Verify
+         env:
+            CERTORAKEY: ${{ secrets.CERTORAKEY }}
+         run: |
+            # Add your code here
```
Roy-Certora, line 58: This is more of @shoham-certora 's concern,
but do we want to explain how to add conf files to the list?

ozcodes: i'm not sure, how do we usually add it ?

Roy-Certora: We just need an example like the one I've sent you.

ozcodes: @shoham-certora could you add an example of running a conf file instead of "# Add your code here" ?

shoham-certora: A simple example would be 
```yaml
        run: |
          certoraRun path_to_conf_file.conf
```
A more complicated one:
```yaml
         - name: Verify ${{ matrix.rule }}
         env:
            CERTORAKEY: ${{ secrets.CERTORAKEY }}
         run: 
            certoraRun  ${{ matrix.rule }}

    strategy:
       fail-fast: false
       max-parallel: 16
       matrix:
          rule:
             - path-to-conf1.conf --rule rule1  # run specific rule
             - path-to-conf1.conf --rule rule2
             - path-to-conf2.conf  # run the entire conf
```

docs/user-guide/ci.rst
```diff
@@ -0,0 +1,74 @@
+.. role:: bash(code)
+   :language: bash
+
+CI Configuration
+================
+
+Follow these steps to configure CI for GitHub Actions on your repository:
+
+* Step 1:
+   Add the Certora CLI key `as a secret`_ to your repository, the secret name is "CERTORAKEY"
```
shellygr, line 10:
```suggestion
   Add the Certora CLI key `as a secret`_ to your repository, the secret name is `CERTORAKEY`
```

docs/user-guide/ci.rst
```diff
@@ -0,0 +1,74 @@
...
+            - name: Install python
+            uses: actions/setup-python@v2
+            with: { python-version: 3.9 }
+
+            - name: Install java
+            uses: actions/setup-java@v1
+            with: { java-version: "11", java-package: jre }
+
+            - name: Install certora cli
+            run: pip3 install certora-cli==7.0.7
```
shellygr, line 49:
```suggestion
            run: pip3 install certora-cli==7.3.0
```

shellygr: let's add a comment to the yml: "It's recommended to pin the latest `certora-cli` version available in https://pypi.org/project/certora-cli/"

docs/user-guide/ci.rst
```diff
@@ -0,0 +1,74 @@
...
+
+            - name: Install solc
+            run: |
+               wget https://github.com/ethereum/solidity/releases/download/v0.8.23/solc-static-linux
+               chmod +x solc-static-linux
+               sudo mv solc-static-linux /usr/local/bin/solc8.23
```
shellygr, line 55: This isn't standard. Why not instruct to use `solc-select`?
We can put this commented out and above it put a comment "If your project depends on compiling with multiple `solc` versions, you can install and differentiate them using these commands"

docs/user-guide/ci.rst
```diff
@@ -0,0 +1,74 @@
...
+            env:  
+               CERTORAKEY: ${{ secrets.CERTORAKEY }}  
+            run: 
+               certoraRun  ${{ matrix.rule }}  
+
+      strategy:  
+         fail-fast: false  
+         max-parallel: 16  
+         matrix:  
+            rule:
```
shellygr, line 67: let's put a link here to `--rule` and `--exclude_rule` docs:
https://docs.certora.com/en/latest/docs/prover/cli/options.html#rule-rule-name
https://docs.certora.com/en/latest/docs/prover/cli/options.html#exclude-rule-rule-name-pattern

docs/user-guide/ci.rst
ozcodes: @shoham-certora @Roy-Certora do you know if using "solc-select" works better for most of the customers than using a specific version ?

docs/user-guide/ci.rst
ozcodes: @shellygr I couldn't use these as a link inside of a code block.
I'm suggesting writing it at the end of the page, for example:

**Further reading**
`--rule`: https://docs.certora.com/en/latest/docs/prover/cli/options.html#rule-rule-name
`--exclude_rule` https://docs.certora.com/en/latest/docs/prover/cli/options.html#exclude-rule-rule-name-pattern

shellygr: I meant not as link but as part of the yml, but that's also ok. Maybe say it's also related to building the matrix

shellygr (commented):
good modulo the things we discusses, happy to hear feedback from others

I feel pretty strongly about solc8.23 potentially confusing users more than helping them, _some_ comment is necessary

docs/user-guide/ci.rst
ozcodes: got it, i added it next to them, let me know if its looks allright

docs/user-guide/ci.rst
ozcodes: @shellygr i also think using solc-select is better, i added it instead of wget and and also added the comment above it

docs/user-guide/ci.rst
```diff
@@ -0,0 +1,72 @@
...
+
+        - name: Install java
+          uses: actions/setup-java@v1
+          with: { java-version: "11", java-package: jre }
+
+        # It's recommended to pin the latest certora-cli version available in https://pypi.org/project/certora-cli/
+        - name: Install certora cli
+          run: pip install certora-cli==7.3.0
+
+        # If your project depends on compiling with multiple solc versions, you can install and differentiate them using these commands.
```
shellygr, line 49: I meant that ypu can keep the original wget commands under a comment for this comment to be relevant. I'd put it below the solc-select installation

ozcodes: oki hope i got it right this time, let me know if it looks allright

shellygr (approved):

PR #252, opened 2024-04-25
Extends and improves tracking sums pattern

shoham-certora:
- Extensive explanations regarding tracking sums.
- One example uses the [Examples repository](https://github.com/Certora/Examples/).
- Removes dead link to tutorials.

[Documentation link](https://docs.certora.com/en/shoham-tracking-sums/docs/user-guide/patterns/sums.html)


docs/user-guide/patterns/sums.rst
```diff
@@ -0,0 +1,69 @@
...
+assumptions. So instead, we shall prove a stronger property, as explained next.
+
+
+Equality of sum of balances and total supply
+--------------------------------------------
+The preferred solution to tracking sums is using a hook and a ghost, as shown below.
+
+.. cvlinclude:: ../../../Examples/DEFI/ERC20/certora/specs/ERC20Fixed.spec
+   :lines: 98-106, 115-116
+   :caption: :clink:`Total supply is sum of balances</DEFI/ERC20/certora/specs/ERC20Fixed.spec>`
```
RedLikeRosesss, line 66: I would add a caption link with specific lines: https://github.com/Certora/Examples/blob/225e3b16f5728d975c4d52a47f5ffa2f877c1c16/DEFI/ERC20/certora/specs/ERC20Fixed.spec#L96-L116

docs/user-guide/patterns/sums.rst
```diff
@@ -0,0 +1,69 @@
...
+
+Once this invariant is proved, we can require properties like the one in
+:cvl:`sumOfTwo` above.
```
RedLikeRosesss, line 69: I think it's not clear why one should come back to `sumOfTwo ` if they have a better invariant `totalSupplyIsSumOfBalances`.

docs/user-guide/patterns/sums.rst
```diff
@@ -0,0 +1,69 @@
...
+We see that the Prover cannot verify :cvl:`sumOfTwo` invariant without us adding unsound
+assumptions. So instead, we shall prove a stronger property, as explained next.
+
+
+Equality of sum of balances and total supply
+--------------------------------------------
+The preferred solution to tracking sums is using a hook and a ghost, as shown below.
+
+.. cvlinclude:: ../../../Examples/DEFI/ERC20/certora/specs/ERC20Fixed.spec
+   :lines: 98-106, 115-116
```
RedLikeRosesss, line 65: `Sload` hook is missing form the example on doc page. not everyone will go to github to copy the whole solution if they don't know about it

docs/user-guide/patterns/sums.rst
```diff
@@ -0,0 +1,69 @@
...
+
+
+Trying to verify the sum of two balances
+----------------------------------------
+Often one needs a invariant like :cvl:`sumOfTwo` below:
+
+.. code-block:: cvl
+
+   invariant sumOfTwo(address a, address b)
+       (a != b) => (balanceOf(a) + balanceOf(b) <= to_mathint(totalSupply()));
```
RedLikeRosesss, line 21: wasn't it a part of a hook in a previous solution we had? Why did you remove it?

docs/user-guide/patterns/sums.rst
```diff
@@ -0,0 +1,69 @@
...
+This section deals with tracking sums of values. The quintessential example for
+needing to track sums is from ERC-20 contracts, where :solidity:`totalSupply()`
+must be the sum of all balances. This is the example we shall use here,
+namely the :clink:`ERC20</DEFI/ERC20/contracts/ERC20.sol>` contract from the
+:clink:`Examples</>` repository.
+
+
+Trying to verify the sum of two balances
+----------------------------------------
+Often one needs a invariant like :cvl:`sumOfTwo` below:
```
RedLikeRosesss, line 16: "a invariant"? I think it should be "an" or maybe even "the' because we talk about specific invariant. But definitely not "a"

I think the transition from the first paragraph to this part isn't clear. We need to check the sum of ALL balances, but now we downgraded our property to the sum of only two balances.
Maybe something like: "First of all, let's start with a simpler property: sum of two balances is less than or equal to `totalSupply()`. It will also improve our understanding of how the prover works and finds counter-examples."

docs/user-guide/patterns/sums.rst
shoham-certora: You're right. The problem is that the `:clink:` command doesn't support lines yet. I'll add it to my TODO list for that component.

RedLikeRosesss:
Looks good now. Should we wait for one more review?

shoham-certora:
> Looks good now. Should we wait for one more review?

@RedLikeRosesss , I also think it looks better now, but I'd like to ask @nd-certora about the `require` in the `Sload` hook - I don't like it.

urikirsh:
@shoham-certora can this PR be merged for the next Prover release?

PR #253, opened 2024-04-25
Uri/updating flag names

urikirsh:
Reflects flag name changes that were shortened. Also removes documentation for the deprecated flag `--use_memory_safe_autofinders` and documents instead `--no_memory_safe_autofinders`.

Link to generated documentation: https://certora-certora-prover-documentation--253.com.readthedocs.build/en/253/

### DIFF
--- a/docs/prover/cli/options.md
+++ b/docs/prover/cli/options.md
@@ -591,24 +591,21 @@ The notification in the rule report that contains the applied summaries will pre
 certoraRun Bank.sol --verify Bank:Bank.spec --nondet_difficult_funcs --nondet_minimal_difficulty 20
 ```
 
-(--use_memory_safe_autofinders)=
-### `--use_memory_safe_autofinders`
+(--no_memory_safe_autofinders)=
+### `--no_memory_safe_autofinders`
 
 **What does it do?**
-This option avoids compilation errors when using `certoraRun` by marking auto-generated instrumented assembly
-code as `memory-safe`.
+This option allows compilation errors when using `certoraRun` by not marking auto-generated 
+instrumented assembly code as `memory-safe`.
 
 **When to use it**
-If you see `solc` failures right after the `certoraRun` prints
-`Compiling XXX to expose internal function information...`,
-and the compiled contract is compiled with Solidity version 0.8.13 and above.
-
-It is usually useful in conjunction with {ref}`--solc_via_ir`.
+When internal functions are not found or internal function summaries are not applied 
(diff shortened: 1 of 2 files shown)


docs/prover/cli/options.md
```diff
@@ -570,45 +570,42 @@ verification results, as well as highlighting potentially difficult functions.
...
 
-(--use_memory_safe_autofinders)=
-### `--use_memory_safe_autofinders`
+(--no_memory_safe_autofinders)=
+### `--no_memory_safe_autofinders`
 
 **What does it do?**
-This option avoids compilation errors when using `certoraRun` by marking auto-generated instrumented assembly
-code as `memory-safe`.
+This option allows compilation errors when using `certoraRun` by not marking auto-generated
```
shellygr, line 598: I would write the inverse way
```suggestion
This option avoids marking auto-generated instrumented assembly code as `memory-safe`, and may thus cause Solidity compilation errors when using `certoraRun`.
```

docs/prover/cli/options.md
```diff
@@ -570,45 +570,42 @@ verification results, as well as highlighting potentially difficult functions.
...
+instrumented assembly code as `memory-safe`.
 
 **When to use it**
-If you see `solc` failures right after the `certoraRun` prints
-`Compiling XXX to expose internal function information...`,
-and the compiled contract is compiled with Solidity version 0.8.13 and above.
-
-It is usually useful in conjunction with {ref}`--solc_via_ir`.
+When internal functions are not found or internal function summaries are not applied 
+and the contract was compiled with Solidity version 0.8.13 and above.
```
shellygr, line 603:
```suggestion
When one wishes to avoid adding `memory-safe` to auto-generated assembly blocks.
```

shellygr (commented):
there's no real reason I see to disable memory-safe assembly blocks for autofinders, so adjusted the text.

we need to put a label on the pr that it is for an upcoming version since these changes are already in master

urikirsh:
If there is no real reason to use this flag, then I think it's best to remove it completely from our documentation and the help message (we already have a lot of flags).
What would be the number of the next version? V 7.6.0?
@shellygr

urikirsh:
@shellygr which label do we put on PRs that are are a part of the coming release?

shellygr:
Let's put a label "release" until we know the version number. And we can define it to mean that it is everything merged to master already and is waiting to enter beta

shellygr (approved):

PR #257, opened 2024-05-23
CERT-6251 - make sure all flags in documentation have a corresponding flag

rahav-certora:
- The motivation for this PR is that in the new dashboard we will see the list of flags used for the run and the user will be able to click on the flag name and get to the online documentation 
- The flag and the anchor should be the same to allow a generic link creation based on flag name

These flags have help message i.e. will appear in certoraRun --help but do not have documentation in certora web documentation.

PROTOCOL_NAME
PROTOCOL_AUTHOR
BUILD_CACHE
MULTI_EXAMPLE
FUNCTION_FINDER_MODE
VYPER
SOLC_EXPERIMENTAL_VIA_IR
SOLC_MAP
STRICT_SOLC_OPTIMIZER
YUL_ABI
TYPECHECKER_ARGS
IGNORE_SOLIDITY_WARNINGS
PROVER_VERSION
BYTECODE_JSONS

### DIFF
--- a/docs/prover/cli/options.md
+++ b/docs/prover/cli/options.md
@@ -31,6 +31,7 @@ Modes of operation
 
 The Certora Prover has three modes of operation. The modes are mutually exclusive - you cannot run the tool with more than one mode at a time.
 
+(--verify)=
 ### <a id="verify"></a>`--verify`
 
 **What does it do?**
@@ -45,7 +46,6 @@ If we have a Solidity file `Bank.sol`, with a contract named `Bank` inside it, a
 
 Most frequently used options
 ----------------------------
-
 
 ### <a id="msg"></a>`--msg <description>`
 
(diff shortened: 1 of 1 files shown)


alexandernutz:
so what do these `<a ..`s you added do technically? Is there any code that looks at them (even in CI?) or is this our due dilligence to keep them in sync?
Also wondering: weren't there implicit achors already, inserted by our markdown-to-html (sphinx or myth or something)?

I guess I'm wondering what our bigger plans are here, and how this fits ..

rahav-certora:
- Original anchors that included dashes were left as is. 
- Combining the anchor with the title is more readable and safer (we had cases of flag renaming without renaming the anchor). Also the use of HTML anchor is accepted by virtually all md interperters

alexandernutz (approved):
looks good to me 
(the reasoning why we want it makes sense, the names seem ok, and in the compiled doc there seems to be no change)
(I dunno if the other reviewers want to look so maybe check with them if you think so, but the change seems non-dangerous, so I'm approving)

PR #260, opened 2024-06-03
CERT-6348: Fix typo, still using `--structLink`

nafur:
We renamed `--structLink` to `--struct_link` some time ago, but this was still using the old one.


shoham-certora (approved):
:+1:  Approved

PR #261, opened 2024-06-11
Adding definition of Persistent Ghost

johspaeth:
In the current version of the documentation we fail to say explicitly what a persistent ghost is. This PR adds a short definition-

See also discussion around https://certora.slack.com/archives/CF2E0UR3L/p1718090375662839

### DIFF
--- a/docs/cvl/ghosts.md
+++ b/docs/cvl/ghosts.md
@@ -194,6 +194,8 @@ Restrictions on ghost axioms
 Ghosts vs. persistent ghosts
 ----------------------------
 
+A `persistent ghost` is a `ghost` that will never be {ref}`havoc <glossary>`. This typically happens for unresolved calls, when the Prover will automatically {ref}`havoc <glossary>` the storage. 
+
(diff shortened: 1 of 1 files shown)


docs/cvl/ghosts.md
```diff
@@ -194,6 +194,8 @@ Restrictions on ghost axioms
 Ghosts vs. persistent ghosts
 ----------------------------
 
+A `persistent ghost` is a `ghost` that will never be {ref}`havoc <glossary>`. The value of a non-persistent `ghost` will be `havoc'ed` when the Prover `havoc`s the storage, a `persistent ghost` however will keep its value when storage is havoced.
```
shoham-certora, line 197: We should also mention that persistent ghosts do not revert.

shoham-certora (approved):
See comment.

docs/cvl/ghosts.md
johspaeth: @shoham-certora Do you mean persistent ghost _functions_ do not revert?

PR #262, opened 2024-06-18
Shoham/small fixes june24

shoham-certora:
## Small fixes

1. Make the behavior of `calledContract` explicit
2. Fix typo in https://docs.certora.com/en/latest/docs/prover/approx/grounding.html#how-grounding-works

### Jira tickets:
- https://certora.atlassian.net/browse/DOC-423
- https://certora.atlassian.net/browse/DOC-424


docs/cvl/methods.md
```diff
@@ -748,8 +748,10 @@ supported.
 The function call may also use the special variable `calledContract`, which
 gives the address of the contract on which the summarized method was called.
 This is useful for identifying the called contract in {ref}`wildcard summaries
-<cvl2-wildcards>`.  The `calledContract` keyword is only defined in the `methods`
-block.
+<cvl2-wildcards>`, both for external and _internal_ functions. Note that for
```
alexandernutz, line 751: why the highlight of "internal"?

docs/cvl/methods.md
```diff
@@ -748,8 +748,10 @@ supported.
...
+library functions the `calledContract` is the contract _calling_ the library
```
Roy-Certora, line 752: And for internal functions which are not in libraries this isn't true?

docs/cvl/methods.md
```diff
@@ -748,8 +748,10 @@ supported.
...
+function.
+The `calledContract` keyword is only defined in the `methods` block.
```
alexandernutz, line 754: maybe "may only be used" (more "common speak" i'd say)

alexandernutz (commented):
nits, ready to approve when other reviewers are ..

docs/cvl/methods.md
shoham-certora: Removed highlight.

docs/cvl/methods.md
shoham-certora: For internal non-library functions it is both the called and calling contract (since they are the same).
Should we state this?

docs/cvl/methods.md
shoham-certora: Done

docs/cvl/methods.md
Roy-Certora: IMO it's worth emphasizing that the `calledContract` is the contract itself. The same is true for library functions (internal library functions are a part of the contract bytecode).

Roy-Certora (approved):

PR #266, opened 2024-06-21
Allow comparison of structs also when they contain bytes or string fields

johspaeth:
When https://github.com/Certora/EVMVerifier/pull/6119 lands in master, we are able to compare bytes and structs.


urikirsh (approved):
lgtm

PR #269, opened 2024-06-21
Updates due to complex types

johspaeth:
With complex types feature we managed to remove the restriction on internal method summaries. We can remove this info from the docs now.


urikirsh (approved):
Sweet! Thanks for updating this!

PR #272, opened 2024-06-28
update for new builtin sanity [DOC-425]

alexandernutz:
What's new is the extra subrule that checks the pessimistic assertions (also tweak description of satisfy a little).

Jira ticket: https://certora.atlassian.net/browse/DOC-425
Link to generated documentation:https://certora-certora-prover-documentation--272.com.readthedocs.build/en/272/docs/cvl/builtin.html#how-sanity-is-checked


docs/cvl/builtin.md
```diff
@@ -106,12 +106,19 @@ rule sanity {
...
 }
 ```
 
-To find a satisfying trace, the Prover must construct an input
-for which `f` doesn't revert.
+This will create two sub-rules, which will be visible in the report. One
+sub-rule checks the `satisfy true` statement, which is fulfilled if there is an
+input such that `f(e, arg)` runs to completion without reverting. The other
+sub-rule checks the `assert true` statement. Of course, this assertion itself is
+never violated, but the sub-rule contains the "pessimistic" assertions that we
```
johspaeth, line 118: We should also state what the pessimistic assertions check for - and roughly explain where they are added.

docs/cvl/statements.md
```diff
@@ -150,6 +152,15 @@ some arbitrary state.  It is not possible to check that every possible starting
 state has an execution that satisfies the condition.
 ```
 
+```{note}
+Rules without any explicit `assert` statements will not check the "pessimistic" 
+assertions, i.e., the implicit assertions that we insert in rules with `assert`
```
johspaeth, line 157: Also here, it'd be better to explain what these assertions check for.

johspaeth (commented):
We should also explicitly add what the "pessimistic" asserts check for and where we insert them when running the builtin sanity rule.

shellygr (commented):
Can you explain the 'right before' part?

docs/cvl/statements.md
```diff
@@ -122,25 +122,27 @@ rule totalFundsAfterDeposit(uint256 amount) {
...
-for each input variable and each state variable where all `require` and `satisfy`
-statements are executed successfully.  In case there is no such execution, for
-example if the `require` statements are already inconsistent or if a solidity
-function always reverts, an error is reported.
+For each `satisfy` statement that is checked successfully, the Certora verifier
+will produce a witness for a valid execution of the rule.  It will show an
+execution trace containing values for each input variable and each state
+variable where all `require` and `satisfy` statements are executed successfully.
+In case there is no such execution, for example if the `require` statements are
+already inconsistent or if a solidity function always reverts, the rule will
```
urikirsh, line 138:
```suggestion
already inconsistent or if a Solidity function always reverts, the rule will
```

urikirsh (commented):
I am writing the copy of the version update email. I read the PR and I don't understand what is the benefit of adding this rule. Is it to help narrow down the cause of a sanity failure (if the new rule fails, the function always reverts regardless of any optimistic assumptions)?

alexandernutz:
> I am writing the copy of the version update email. I read the PR and I don't understand what is the benefit of adding this rule. Is it to help narrow down the cause of a sanity failure (if the new rule fails, the function always reverts regardless of any optimistic assumptions)?

Before this change, `builtin sanity` was behaving as if all optimistic options were set, no matter their actual status. In practice, this resulted in the sanity rule sometimes failing due to e.g. hashing or loop bounds being exceeded on all inputs without giving users a hint about this. Since those optimistic assumptions are implicit, it was rather hard to understand those cases of vacuity. After the change, in these cases there will be an extra sub-rule telling the user about the pessimistic assertions being violated, so they have an easier time tracking down the vacuity. (The technical background is that `satisfy` rules are always "optimistic", and `builtin sanity` had no `assert`, so the solution was to add an `assert true` to the rule.) 

Here's an example
https://certora.atlassian.net/issues/CERT-6114

docs/cvl/statements.md
alexandernutz: changing

docs/cvl/builtin.md
alexandernutz: hm, the difficulty is that we use the optimistic/pessimistic principle in a whole list of instances (see the CLI options page -- I just checked and saw things I had never heard about 🙃 ) 

One more thing to consider is that I don't want to write something that needs changing whenever one of those options changes..

Best thing I can think of right now is making a glossary entry `pessimistic assertions` (with an alias `optimistic assumptions`) that explains the principle and lists some of the options.. wdyt?

alexandernutz:
> Can you explain the 'right before' part?

The "right before" isn't so important.

Changed to
"In order to trigger creation of a sub-rule that contains these checks, users
can insert an `assert true` statement into the rule. This assert itself will 
never be violated, but the sub-rule we create for it will contain all the 
"pessimistic" assertions for the program."

is that better?..

.. at leat for now .. I still have to get to Johannes' request and properly explain what the pessimistic assertions are ..

docs/cvl/builtin.md
alexandernutz: alright, did it and re-requested -- what do you think @johspaeth ? (re-requested)

alexandernutz:
made changes, re-requesting all around ..

chandrakananandi (approved):
just looked at the gambit part.

chandrakananandi (dismissed):
Seems like i cant change my review?

alexandernutz:
> Seems like i cant change my review?

alexandernutz:
> > Seems like i cant change my review?

re-requested .. hm, I guess I can use `prover-cli-options` ... or can we rename away the gambit option some other way?

sorry for changing gambit.md, now I vaguely remember a message about that ...

alexandernutz:
@chandrakananandi I evaded the problem by undoing the gambit.md changes and instead renaming the flag on the prover side

chandrakananandi:
> @chandrakananandi I evaded the problem by undoing the gambit.md changes and instead renaming the flag on the prover side

Ah ok, sorry about the hassle. I slacked you more details but let me know if you want me to change this in the gambit.md file in the gambit repo. I can do it right now!

alexandernutz:
> > @chandrakananandi I evaded the problem by undoing the gambit.md changes and instead renaming the flag on the prover side
> 
> Ah ok, sorry about the hassle. I slacked you more details but let me know if you want me to change this in the gambit.md file in the gambit repo. I can do it right now!

nah .. `prover-cli-options` is an ok name for the label as well :-) 
(generally, prefixing `gambit-` to those labels with very generic names might be a consideration .. but then, we didn't do it for the prover either ..)

johspaeth (approved):
With the additional explanation of pessimistic assertions the changes make sense to me now. Thanks.

PR #274, opened 2024-07-01
glossary entry for call trace (was: TODO)

alexandernutz:
In the glossary, the "call trace" entry showed "TODO". Adding an actual entry.

Link to generated documentation: https://certora-certora-prover-documentation--274.com.readthedocs.build/en/274/docs/user-guide/glossary.html#term-call-trace


pickx (approved):
looks great
(tioli: maybe start with saying that this is only generated when violations/satisfies are found, since the common case is no call trace)

nd-certora (approved):

PR #276, opened 2024-07-07
Remove duplicate descriptions of the `--method` flag

naftali-g:
Discovered while working on CERT-6639


docs/prover/cli/options.md
```diff
@@ -647,11 +647,7 @@ Options that help reduce the running time
 
 ### `--method`
 
-**What does it do?**
-Parametric rules will only verify the method with the given signature, instead of all public and external methods of the contract. Note that you will need to wrap the method's signature with quotes, as the shell doesn't interpret parenthesis correctly otherwise.
-
-**When to use it?**
-When you are trying to solve/understand a counterexample of a parametric rule on a specific method.
+See {ref}`--method`
```
yoav-el-certora, line 650: The only issue imo is that both here and in the [--method](https://docs.certora.com/en/latest/docs/prover/cli/options.html#method) option, the example we provide is string and not a list.

naftali-g: added an example of command-line multiple methods. For the conf case it;s just like all the other flags we have that support lists, I don;t think we need any special docs for it

docs/prover/cli/options.md
```diff
@@ -647,11 +653,7 @@ Options that help reduce the running time
...
 
 **Example**
```
urikirsh, line 658: Not sure if this is needed - there is an example in the link, and better to not keep several references which might get out of sync

urikirsh (approved):

docs/prover/cli/options.md
naftali-g: Yah I missed this. Removing it as well.

docs/prover/cli/options.md
```diff
@@ -156,6 +156,12 @@ the contract or the spec to rerun, we can just rerun on the `deposit` method:
 certoraRun --method 'deposit(uint)'
 ```
 
+If we discover a counterexample in several methods, we could rerun just those:
+
+```sh
+certoraRun --method 'deposit(uint)' --method 'transfer(address,uint256)'
```
johspaeth, line 162: @naftali-g shouldn't this be either
```suggestion
certoraRun --method ['deposit(uint)', 'transfer(address,uint256)']
```

johspaeth (changes requested):
With the changes (that have landed in https://github.com/Certora/EVMVerifier/pull/6164) - I think we need to update some of the examples - see comment below?

docs/prover/cli/options.md
naftali-g: No, the way to specify multiple methods from command line is with multiple usages of the `--method` flag. In the conf file it's with square brackets, like any list.

johspaeth: Thanks for clarification!

johspaeth (approved):
Approving after clarification.

PR #277, opened 2024-07-08
CERT-6671 - fix anchors

rahav-certora:
The anchors to the flag documentation do not work correctly for example `https://docs.certora.com/en/latest/docs/prover/cli/options.html#method`

goes to the table of content not to the paragraph itself. 
Fix by the following:

- using only the (XXX)= format
- underscores are not supported in anchors (changed automatically to dashes) 
- anchors should be in lowercase - toggled case when needed


shoham-certora (approved):
:+1:  Approved

PR #264, opened 2024-06-21
Improving invariant section to state which function are being checked

johspaeth:
Naming convention:
 - PRs for features that are in design should have the "proposal" label
 - PRs for features that haven't landed yet should have the "future" label
 - PRs for upcoming releases should have the "release" label
 - PRs with new documentation for existing features should have the "existing feature" label

Before requesting review:
 - Make sure there is a ticket in the DOC board in Jira
 - Make sure CI is passing
   - Spell check failure may require adding backticks around code or updating `spelling_wordlist.txt`
   - See `README.md` for information about style and markdown syntax
   - If the CI Details link gives a 404, you need to log in to readthedocs.com
 - Add link to generated documentation here
   - you can find this by following the read the docs link from the CI check
 - Ask for help in #documentation

Jira ticket: TODO
Link to generated documentation: TODO

PR #278, opened 2024-07-09
Documentation for Strong Invariants

johspaeth:
This PR adds documentation for the upcoming strong invariant feature that will be part of beta version `>7.9.0`. 

To avoid merge conflicts, I drafted it on top of [this PR](https://github.com/Certora/Documentation/pull/264) which will be merged for the main release of version `7.9.0`


docs/cvl/invariants.md
```diff
@@ -75,6 +78,25 @@ and by {ref}`--parametric_contracts`. `View` and `pure` methods are excluded fro
...
+
+ 1. Assert that the invariant holds before the `c` - if the invariant does not hold due to some logic of the current method, this will yield a counter example.
+ 2. Assume the invariant holds after the call `c`. The semantics is that the call did not break the invariant.  
+ 3. In the case `c` is a `delegatecall` havoc the current's contact storage and assert the invariant once more. 
+
+The havocing logic in step 3 models the scenario that a `delegatecall` modifies the current contract's storage.
+
+```{note}
+An invariant that is neither declared as `strong` nor `weak` will be treated as a `weak` invariant by default. 
+This behavior can be changed using the flag `--prover_args '-defaultInvariantType strong'`.
```
urikirsh, line 96: I don't think we should mention this flag:
1. Adds complexity
2. It means we will now have to officially support this prover args syntax
3. If it is a useful option, it should be a flag - flags have better validation, better error messages, and they appear in the help message.
4. The workaround is rather easy (just add `strong` when relevant)

docs/cvl/invariants.md
```diff
@@ -75,6 +78,25 @@ and by {ref}`--parametric_contracts`. `View` and `pure` methods are excluded fro
...
+A strong invariant performs the same checks as a weak invariant - i.e. it will be checked for the constructor
+and for any other method it will be assumed before executing the method (pre-state) and asserted afterward execution of the method (post-state). 
+In addition to these steps, a strong invariant also asserts and assumes the invariant _during_ method
+execution at location that potentially break the invariant. The invariant can be violated if there
+is an unresolved external call that is able to modify the state of the current contract. To verify the strong invariant, for every unresolved external call `c` 
+(a call that will force the prover to havoc storage), a strong invariant will do the following steps:
+
+ 1. Assert that the invariant holds before the `c` - if the invariant does not hold due to some logic of the current method, this will yield a counter example.
+ 2. Assume the invariant holds after the call `c`. The semantics is that the call did not break the invariant.  
+ 3. In the case `c` is a `delegatecall` havoc the current's contact storage and assert the invariant once more.
```
urikirsh, line 90: Not clear to me what exactly happens and why we do this. Does 3 replace step 2, or happens right after it?
Is there any realistic case where there is an invariant we can prove right after a havoc? (AFAIK resetting storage also resets ghosts)

docs/cvl/invariants.md
```diff
@@ -55,16 +55,19 @@ production, and {doc}`statements` for the `block` production.
...
 --------
 
-In CVL, an invariant is a property of the contract state that is expected to be
+In CVL, we distinguish between strong and weak invariants.
+A *weak* invariant is a property that is expected to be
 true whenever a contract method is not currently executing.  This kind of
-invariant is sometimes called a "representation invariant".
+invariant is sometimes called a "representation invariant". A *strong* invariant is 
+an invariant that also holds before and after execution of an unresolved call, i.e. a call that
+potentially calls to another contract which could modify the current contract's state.
```
urikirsh, line 63: This paragraph still doesn't explain when should we use strong invariants over weak invariants. Perhaps we should base on the first sentence from the example at https://github.com/Certora/Examples/blob/cli-beta/CVLByExample/StrongInvariants/README.md:

"Strong invariants are sure to hold whenever control is yielded to an external function, providing enhanced security, especially for contracts without global locks. However, they add additional checks that can result in longer run times or even halted jobs."

urikirsh: In general, linking to the example would be highly beneficial

docs/cvl/invariants.md
jhoenicke: Step 3 is in addition and should provide soundness.  A delegatecall can change storage, so we simulate this by havocing storage.  We do this at the end, because we need to show that even after step 2, the delegator cannot break the invariant (we don't need it then after step 1, because the solver can always assume that step 2 does no change, so only checking after step 2 also catches the bug).

I think step3 does not havoc ghost (is this right?)  The idea is that the delegator cannot change the ghost state or the state of any other contract, just the contract that it is called on.  

This is mostly for invariants that don't even talk about the state of the contract. If the contract that does the delegate call is not the main contract we want to reason about, then delegatecalls from this contract are always okay.

docs/cvl/invariants.md
jhoenicke: I can agree with not mentioning it, (but we should have a list of flags for ourselves)

But the idea was to keep the flag so that we can gather more experience, whether strong or weak is the better default.

docs/cvl/invariants.md
```diff
@@ -55,16 +55,19 @@ production, and {doc}`statements` for the `block` production.
...
+an invariant that also holds before and after execution of an unresolved call, i.e. a call that
+potentially calls to another contract which could modify the current contract's state. 
 
 Each invariant has a name, possibly followed by a set of parameters, followed
 by a boolean expression.  We say the invariant *holds* if the expression
 evaluates to true in every reachable state of the contract, and for all
 possible values of the parameters.
 
-While verifying an invariant, the Prover checks two things.  First, it checks
+While verifying an weak invariant, the Prover checks two things.  First, it checks
```
jhoenicke, line 70:
```suggestion
While verifying a weak invariant, the Prover checks two things.  First, it checks
```

docs/cvl/invariants.md
johspaeth: Regarding flags for internal use:

I bookmarked these two links:
* https://github.com/Certora/EVMVerifier/blob/master/lib/Shared/src/main/kotlin/config/Config.kt
* https://github.com/Certora/EVMVerifier/blob/master/scripts/EVMVerifier/certoraContextAttribute.py

docs/cvl/invariants.md
johspaeth: Thanks - I linked the example and improved the wording.

docs/cvl/invariants.md
johspaeth: > I think step3 does not havoc ghost (is this right?)

Correct. 

I updated the description to be more meaningful. Is it better now?

docs/cvl/invariants.md
```diff
@@ -75,6 +79,19 @@ and by {ref}`--parametric_contracts`. `View` and `pure` methods are excluded fro
 invariant checking as by definition they cannot change the state 
 of their contract. 
 
+A strong invariant performs the same checks as a weak invariant - i.e. it will be checked for the constructor
+and for any other method it will be assumed before executing the method (pre-state) and asserted afterward execution of the method (post-state). 
+In addition to these steps, a strong invariant also asserts and assumes the invariant _during_ method
+execution at location that potentially break the invariant. The invariant can be violated if there
```
urikirsh, line 85:
```suggestion
execution at locations that potentially break the invariant. The invariant can be violated if there
```

docs/cvl/invariants.md
```diff
@@ -75,6 +79,19 @@ and by {ref}`--parametric_contracts`. `View` and `pure` methods are excluded fro
 invariant checking as by definition they cannot change the state 
 of their contract. 
 
+A strong invariant performs the same checks as a weak invariant - i.e. it will be checked for the constructor
+and for any other method it will be assumed before executing the method (pre-state) and asserted afterward execution of the method (post-state).
```
urikirsh, line 83:
```suggestion
and for any other method, it will be assumed before executing the method (pre-state) and asserted afterward execution of the method (post-state). 
```

docs/cvl/invariants.md
```diff
@@ -75,6 +79,19 @@ and by {ref}`--parametric_contracts`. `View` and `pure` methods are excluded fro
 invariant checking as by definition they cannot change the state 
 of their contract. 
 
+A strong invariant performs the same checks as a weak invariant - i.e. it will be checked for the constructor
+and for any other method it will be assumed before executing the method (pre-state) and asserted afterward execution of the method (post-state). 
+In addition to these steps, a strong invariant also asserts and assumes the invariant _during_ method
+execution at location that potentially break the invariant. The invariant can be violated if there
+is an unresolved external call that is able to modify the state of the current contract. To verify the strong invariant, for every unresolved external call `c`
```
urikirsh, line 86:
```suggestion
is an unresolved external call that can modify the state of the current contract. To verify the strong invariant, for every unresolved external call `c` 
```

urikirsh (approved):
Small suggestions

PR #279, opened 2024-07-10
Remove old documentation

shoham-certora:
## Remove old documentation
1. Removes the old documentation `docs/confluence/`.
2. Fixes references to the old documentation.
3. Enables seeing complete TODO list in local builds.


alexandernutz (commented):
hmm .. I'm not sure about this -- feels like we need a strategy decision..

From a rough scan, maybe we can categorize the old docs like this
 - still applicable and we have a replacement
 - still applicable and we don't have a replacement
 - out of date / wrong / useless 

--> I don't know if anyone has the time, but it might be wise to sort through the old docs with that in mind ..

to mitigate the current status with low effort, we could
 - delete them now, restore them later (basically go through with this PR)
 - leave as is, but make a more prominent warning in some way (that was asked in discord I understand ..)
    - if we leave them, it would be really good if they would also be marked in the search (prioritized lower, highlighted ...) - I suspect it does make for a bad UI if they clog the search results 
  
(also, some misc comments)

docs/confluence/pitfalls.md
```diff
@@ -1,98 +0,0 @@
...
-
-### Raw arguments (soon to be solved!)
-
-The `calldataarg` type represents the full byte array passed as `calldata` to
-the EVM bytecode. It includes, in particular, the 4-byte sighash used by the
-ABI specification to identify the high-level function executed. Therefore, the
-following spec will pass vacuously:
-
-```cvl
-rule dontForgetThe4Bytes {
```
shellygr, line 50: this is still relevant though

docs/confluence/perplexed.md
```diff
@@ -1,166 +0,0 @@
...
-![](attachments/249724929/249790465.png)
-
-As you can see there’s a ~2 hours difference between “Queue Time” and “Finish Time”.
-
-**Action items:**
-
-1.  Figure out if the problem happened due to having too many rules and SMT taking too long, or if preprocessing has failed to complete in time. Therefore, open the linked-to **zip file**, extract it, and look into `Reports/Results.txt`. If there is no message “Has CVL checks” or “Starting to run on rule ….” then there is a good chance that preprocessing took too long.  
-    If you did not even see a message “The scene contains … contracts” it means that even the first step of preprocessing took too long.
-    
-2.  If the problem is preprocessing, first try to run with `-enableEqualitySaturation=false`, and if that does not help try in addition `-simplificationDepth=10`. If both don't help, please report to Certora.
```
shellygr, line 30: this tip may be still relevant at times

shellygr (approved):
2 nits

index.rst
```diff
@@ -59,6 +58,17 @@ on our `Help Desk channel on Discord`_.
...
+.. Adding TODO list visible only in dev-build.
+   To create a dev-build locally run:
+   sphinx-build -b html . build/html -t is_dev_build
+
+.. only:: is_dev_build
+
+   To do list
+   ----------
+
+   .. todolist::
```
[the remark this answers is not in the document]
shoham-certora: I think this change is purely technical:

* The `todolist` refers to Sphinx todo directives that exist in our code. For example in [Tracking changes with ghosts and hooks](https://github.com/Certora/Documentation/blob/master/docs/user-guide/ghosts.md?plain=1#L24-L28) (this todo comment is from September) see image below.
* Before February these were shown in our Documentation. Since then they are omitted from the build, but can be recovered in a local builds using `-t is_dev_build` argument.
* The current change just adds a list of all these todo comments in the landing change, when building locally with  `-t is_dev_build` argument.

image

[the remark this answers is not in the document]
shoham-certora: Personally I think these todo comments are superior to Jira tasks: 
1. In a todo comment I know at least which file and what lines it refers to, (as opposed to various Jira tasks that only have an obscure title).
2. Whenever anyone changes the relevant file, there is a chance they will also resolve the todo comment. As opposed to old Jira tasks that fester at the bottom of the backlog.

docs/confluence/pitfalls.md
shoham-certora: Should I add these as part of this PR or do it later?

PR #285, opened 2024-08-02
Small fix in example

christiane-certora:
Add missing == 0 in rewritten example about quantified variables as arguments


urikirsh (approved):
Approving and merging as an effort to reduce the backlog of Doc PRs

PR #290, opened 2024-08-30
add ASSERT_FALSE summaries

christiane-certora:
I hope I captured the purpose of these summaries correctly, happy to change it if someone more involved with the change has something else to mention.

### DIFF
--- a/docs/cvl/methods.md
+++ b/docs/cvl/methods.md
@@ -68,6 +68,7 @@ method_summary   ::= "ALWAYS" "(" value ")"
                    | "HAVOC_ALL"
                    | "DISPATCHER" [ "(" ( "true" | "false" ) ")" ]
                    | "AUTO"
+                   | "ASSERT_FALSE"
                    | id "(" [ id { "," id } ] ")" [ "expect" id ]
                    | "DISPATCH" "[" dispatch_list_pattern [","] | empty "]" "default" method_summary
 
@@ -448,6 +449,8 @@ There are several kinds of summaries available:
    or {ref}`ghost-axioms`.
 
  - {ref}`auto-summary` are the default for unresolved calls.
+   
+ - {ref}`assert-false-summary`. These replace the method with an assert false, effectively checking that no such method is called.
 
 (delete-summary)=
 ### Summary application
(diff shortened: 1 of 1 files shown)


docs/cvl/methods.md
```diff
@@ -706,6 +709,13 @@ The behavior of the `AUTO` summary depends on the type of call[^opcodes]:
   [State Mutability](https://docs.soliditylang.org/en/v0.8.12/contracts.html#state-mutability)
   in the Solidity manual for details.
 
+(assert-false-summary)=
+#### `ASSERT_FALSE` summaries
+
+This summary essentially just checks that the summarized method is not reached, 
+which can be useful in the presence of unresolved calls that can be dispatched in many ways.
+It can enable more optimizations in the prover and lead to shorter running times.
```
shellygr, line 717:
```suggestion
This summary checks that the summarized method is not reached, by triggering an assertion failure if a call site to the summarized function is reached.
It can be particularly useful in combination with dispatch list summaries as its default case---to safely rule out calls to methods that do not appear in the dispatch list. In certain cases it can help for optimizations that shorten running times.
```

shellygr (approved):
rephrased a bit, but happy to hear another opinion

docs/cvl/methods.md
johspaeth, line 717: Maybe?

```suggestion
This summary is a short syntax for a summary that contains an `assert false;` and essentially just checks that the summarized method is not reached.
This can be useful for instance, in the presence of unresolved calls in combination with the `unresolved external` syntax to ensure that every unresolved call is actually dispatched correctly (i.e. use `unresolved external in _._ => DISPATCH [...] default ASSERT_FALSE`). It also enables more optimizations in the prover and may lead to shorter running times.
```

johspaeth (commented):
Minor suggestion.

christiane-certora:
Both nice ways, I applied Johannes', since it seemed nice to have the full example for how it would look like as the default on the unresolved dispatcher, but no strong opinion from my side, really.

shellygr:
right but @christiane-certora try to keep the wording short and void of words that do not add meaning such as "essentially" and "just". Also capital letter "Prover"

PR #291, opened 2024-08-30
General expression summaries

christiane-certora:
A very minimal way to not lie about the fact that we can have arbitrary expressions and not just function calls in the summaries. Multiple people seemed recently confused about that.

We could probably write more about it, but if the most prominent usecase still is to just call a function, I'm not sure we have to.

### DIFF
--- a/docs/cvl/methods.md
+++ b/docs/cvl/methods.md
@@ -431,7 +431,7 @@ After the `optional` annotation, an entry may contain a `with(env e)` clause.
 The `with` clause introduces a new variable (`e` for `with(env e)`) to represent
 the {ref}`environment <env>` that is passed to a summarized function; the
 variable can be used in function summaries.  `with` clauses may only be used if
-the entry has a function summary. See {ref}`function-summary` below for more
+the entry has a function summary. See {ref}`expression-summary` below for more
 information about the environment provided by the `with` clause.
 
 
--- a/docs/cvl/methods.md
+++ b/docs/cvl/methods.md
@@ -68,7 +68,7 @@ method_summary   ::= "ALWAYS" "(" value ")"
                    | "HAVOC_ALL"
                    | "DISPATCHER" [ "(" ( "true" | "false" ) ")" ]
                    | "AUTO"
-                   | id "(" [ id { "," id } ] ")" [ "expect" id ]
+                   | expr [ "expect" id ]
                    | "DISPATCH" "[" dispatch_list_pattern [","] | empty "]" "default" method_summary
 
 dispatch_list_patterns ::= dispatch_list_patterns "," dispatch_pattern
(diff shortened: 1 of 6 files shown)


docs/cvl/methods.md
```diff
@@ -706,12 +706,12 @@ The behavior of the `AUTO` summary depends on the type of call[^opcodes]:
   [State Mutability](https://docs.soliditylang.org/en/v0.8.12/contracts.html#state-mutability)
   in the Solidity manual for details.
 
-(function-summary)=
-#### Function summaries
+(expression-summary)=
+#### Expression summaries
```
johspaeth, line 710: I'd leave the change minimal.
```suggestion
(function-summary)=
#### Function and Expression summaries
```

I think we want to leave the focus that `CVL functions` are used as summaries. From a spec writing perspective this is the cleaner solution.

docs/cvl/methods.md
```diff
@@ -444,7 +444,7 @@ There are several kinds of summaries available:
  - {ref}`dispatcher` assume that the receiver of the method call could be any
    contract that implements the method.
 
- - {ref}`function-summary` replace calls to the summarized method with {doc}`functions`
+ - {ref}`expression-summary` replace calls to the summarized method with a CVL expression, typically {doc}`functions`
```
johspaeth, line 447:
```suggestion
 - {ref}`function-summary` replace calls to the summarized method with a CVL expression, typically {doc}`functions`
```

docs/cvl/methods.md
```diff
@@ -706,12 +706,12 @@ The behavior of the `AUTO` summary depends on the type of call[^opcodes]:
   [State Mutability](https://docs.soliditylang.org/en/v0.8.12/contracts.html#state-mutability)
   in the Solidity manual for details.
 
-(function-summary)=
-#### Function summaries
+(expression-summary)=
+#### Expression summaries
 
-Contract methods can also be summarized using CVL {doc}`functions` or
+Contract methods can also be summarized using CVL expressions, typically {doc}`functions` or
```
johspaeth, line 712: Maybe?

```suggestion
Contract methods can also be summarized using CVL functions or expressions (for instance {ref}`ghost-axioms` as approximations).
```

johspaeth (commented):
Thanks for the suggestion.

I'd like to hear @naftali-g take on this as well.

christiane-certora:
I like your version, Johannes, but will wait for Naftali to give his opinion as well first.

docs/cvl/methods.md
naftali-g: I'm actually fine with changing the refname as well.

naftali-g (approved):
definitely an improvement :)

PR #292, opened 2024-09-01
Document address function calls

naftali-g:
Naming convention:

Before requesting review:
 - Make sure CI is passing

Jira ticket: TODO

### DIFF
--- a/docs/cvl/expr.md
+++ b/docs/cvl/expr.md
@@ -369,9 +369,9 @@ Another possible way to have the Prover consider options for a function is by
 prefixing the function name with an `address` typed variable. In this case the
 Prover will consider every possible contract in the scene that implements a
 function that matches the signature provided by the call (if no such function
-exists in the scene the prover will fail with an error). If there are no
+exists in the scene the Prover will fail with an error). If there are no
 constraints on the `address` variable that prevent the Prover from choosing a
-value that doesn't match the address of any contract in the scene, the prover
+value that doesn't match the address of any contract in the scene, the Prover
 will also consider this case and will consider it as a violation of the rule.
 
 (with-revert)=
--- a/docs/cvl/expr.md
+++ b/docs/cvl/expr.md
@@ -108,7 +108,7 @@ Struct Comparison
 CVL supports equality comparison of structs under the following restrictions:
 
  * The structs must be of the same type.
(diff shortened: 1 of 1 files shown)


docs/cvl/expr.md
```diff
@@ -365,6 +365,15 @@ while verifying the rule, and will provide a separate verification report for
...
 {term}`parametric rule`s.
 
+Another possible way to have the Prover consider options for a function is by
+prefixing the function name with an `address` typed variable. In this case the
+Prover will consider every possible contract in the scene that implements a
+function that matches the signature provided by the call (if no such function
+exists in the scene the prover will fail with an error). If there are no
+relevant constraints on the `address` variable the prover may also consider the
+case that its value does not match any contract in the scene and in this case
+the Prover will consider this a violation of the rule.
```
johspaeth, line 375: Can we add an example?

```suggestion
Another possible way to have the Prover consider options for a function is by
prefixing the function name with an `address` typed variable. For instance,
you can write a rule such as `rule r (env e; address a){assert a.balanceOf(e) > 0;}`.
In this case the Prover will consider every possible contract in the scene that 
implements a function that matches the signature provided by the call 
(if no such function exists in the scene the prover will fail with an error). If there 
are no relevant constraints on the `address` variable the prover may also consider 
the case that its value does not match any contract in the scene and in this case
the Prover will consider this a violation of the rule.
```

johspaeth (changes requested):
Can we please add a minimal example here?

johspaeth (approved):
Ok to merge. We'll _not_ include an example for now.

docs/cvl/expr.md
```diff
@@ -365,6 +365,15 @@ while verifying the rule, and will provide a separate verification report for
 each checked method.  Rules that use this feature are referred to as
 {term}`parametric rule`s.
 
+Another possible way to have the Prover consider options for a function is by
+prefixing the function name with an `address` typed variable. In this case the
+Prover will consider every possible contract in the scene that implements a
+function that matches the signature provided by the call (if no such function
+exists in the scene the prover will fail with an error). If there are no
+relevant constraints on the `address` variable the prover may also consider the
```
jar-ben, line 373: _"the prover may also consider"_? Either it considers it or not, do not use "may"

jar-ben, line 373: _"If there are no relevant constraints"_. What do you mean by relevant constraints?

PR #299, opened 2024-09-17
Document address function calls

naftali-g:
This is a ressurection of https://github.com/Certora/Documentation/pull/292, only changing the note about the default behavior (it was `assert false` and now it's `require false`)

Naming convention:

Before requesting review:
 - Make sure CI is passing

Jira ticket: TODO

### DIFF
--- a/docs/cvl/expr.md
+++ b/docs/cvl/expr.md
@@ -108,7 +108,7 @@ Struct Comparison
 CVL supports equality comparison of structs under the following restrictions:
 
  * The structs must be of the same type.
- * The structs (or any nested structs) don't contain dynamic arrays. (`string` and `bytes` can be part of the struct) 
+ * The structs (or any nested structs) don't contain dynamic arrays. (`string` and `bytes` can be part of the struct)
  * There's no support for comparison for structs fetched using direct-storage-access.
 
 Two structs will be evaluated as equal if and only if all the fields are equal.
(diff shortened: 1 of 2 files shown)


docs/cvl/expr.md
```diff
@@ -365,6 +365,15 @@ while verifying the rule, and will provide a separate verification report for
 each checked method.  Rules that use this feature are referred to as
 {term}`parametric rule`s.
 
+Another possible way to have the Prover consider options for a function is by
+prefixing the function name with an `address` typed variable. In this case the
+Prover will consider every possible contract in the scene that implements a
```
shellygr, line 370: do we have a glossary item for `scene`? we should refer to it

docs/cvl/expr.md
```diff
@@ -365,6 +365,15 @@ while verifying the rule, and will provide a separate verification report for
...
+function that matches the signature provided by the call (if no such function
+exists in the scene the prover will fail with an error).
+Note: The Prover will ignore possible values of the `address` variable that
```
shellygr, line 373: maybe we can say it positively:
`The values that the `address` variable can take are the addresses that are associated with the relevant contracts in the scene. Notably, other values would not be possible: Given an address variable `a`, on which we call a some method implemented by contracts `A` and `B`, we will have an implicit `require a == A || a == B`

shellygr (commented):
nits

docs/cvl/expr.md
```diff
@@ -365,6 +365,15 @@ while verifying the rule, and will provide a separate verification report for
 each checked method.  Rules that use this feature are referred to as
 {term}`parametric rule`s.
 
+Another possible way to have the Prover consider options for a function is by
+prefixing the function name with an `address` typed variable. In this case the
```
johspaeth, line 369: Maybe this is slightly simpler wording? Prefixing the function name with `address` type variable reads a bit overly complicated? 

```suggestion
Another possible way to have the Prover consider options for a function is by
using an `address` typed variable and "calling" the function on it, e.g. `address a; a.foo(...);`. In this case the
```

johspaeth (commented):
Minor suggestion

johspaeth (approved):
Fine for me now! Thanks.

PR #313, opened 2024-11-12
Typo Update mutation-verifier.md

mdqst:
**There is a typo in this doc:**

In the section "Randomly generated mutations via Gambit," the sentence:

> To **generated** random mutations via {doc}`Gambit <gambit>`, 

Should be:

> To **generate** random mutations via {doc}`Gambit <gambit>`, 

The word "generated" replaced with "**generate**," as it should be in the infinitive form following "To."

I’m practically a fan of Certora at this point! I admire how they’ve structured their documentation and tools, but even the best of us can slip up occasionally 😄


johspaeth (approved):
Thanks a lot for the PR and improve our documentation, we appreciate it.

PR #318, opened 2024-11-29
Docs: in Whitepaper, Demo link no longer exists

sumitvekariya:
Instead of demo link https://demo.certora.com, which no longer exists, direct readers to [examples](https://github.com/Certora/Examples) repo


urikirsh (approved):
Great fix

PR #319, opened 2024-12-11
CERT-7605: Fix documentation about CVL modulo

nafur:
We had some discussions about what the semantics of the CVL modulo should be. The result is to keep the currently *implemented* semantics, which is different from what our documentation says.


johspaeth (approved):
Thanks for updating. Looks good to me.

PR #335, opened 2025-01-20
Added solana cli options

yoav-el-certora:
Naming convention:

Before requesting review:
 - Make sure CI is passing

Jira ticket: https://certora.atlassian.net/browse/CERT-7992

### DIFF
--- a/docs/solana/options.rst
+++ b/docs/solana/options.rst
@@ -1,5 +1,201 @@
-Solana specific options / CLI flags
-====================================
+Solana-Specific Options / CLI Flags
+===================================
 
-This page documents Solana-specific Certora Prover options, these are either CLI flags or `prover_args` flags.
+This page documents Solana-specific Certora Prover options, which include CLI flags or `prover_args` flags.
 
+The `certoraSolanaProver` utility invokes the Rust compiler and then sends the job to Certora's servers.
+
+The most commonly used command is:
+
+.. code-block:: bash
+
+    certoraSolanaProver --build_script <path_to_build_script>
+
+If a precompiled execution is desired, the run command can skip the compilation step by executing:
+
+.. code-block:: bash
+
+    certoraSolanaProver <path_to_binary_file>
+
+A short summary of these options can be seen by invoking:
+
+.. code-block:: bash
+
+    certoraSolanaProver --help
+
+Using Configuration (Conf) Files
+--------------------------------
+
+For larger projects, the command line for running the Certora Prover can become large and cumbersome. It is therefore recommended to use configuration files instead.
+
+These files are in [JSON5](https://json5.org/) format and use a `.conf` extension. They hold the parameters and options for the Prover.
+
+For more details, see `Conf File <https://docs.certora.com/en/latest/docs/prover/cli/conf-file-api.html#conf-files>`_
+
+
+.. contents:: Overview
+   :depth: 2
+   :local:
+
+Modes of Operation
+------------------
+
+The Certora Solana Prover has two modes of operation. These modes are mutually exclusive - you cannot run the tool with more than one mode at a time.
+
+.. _build_script:
+--build_script
+~~~~~~~~~~~~~~~
+
+**What does it do?**
+
+Runs formal verification of specified properties while providing an automatic method to compile a Rust project.
+
+The build script must output the following:
+
+- `project_directory`: Path to the project root directory.
+- `sources`: List of files or directories used or imported in the program.
+- `executables`: List of compiled binary files, which are the target of the Rust program.
+- `success`: Boolean flag indicating if the build phase passed successfully.
+
+**When to use it?**
+
+Use this mode to prove properties on source code while providing an automatic compilation method. This is especially useful during development when files are modified frequently.
+
+**Example**
+
+.. code-block:: bash
+
+    certoraSolanaProver --build_script <path_to_build_script>
+
+precompiled binary
+~~~~~~~~~~~~~~~~~~~
+
+**What does it do?**
+
+Runs formal verification of specified properties on a precompiled Rust project by providing a path to the binary target file.
+
+**When to use it?**
+
+Use this mode to prove properties on source code without recompiling the project for every execution. Ideal when files are stable and unchanged.
+
+**Example**
+
+.. code-block:: bash
+
+    certoraSolanaProver <path_to_binary_file>
+
+Most Frequently Used Options
+----------------------------
+.. _solana_inlining:
+--solana_inlining
+~~~~~~~~~~~~~~~~~~
+
+**What does it do?**
+
+Provides the prover with a list of paths to inlining files for Solana contracts. These files are parsed and used to prove properties.
+
+**When to use it?**
+
+TODO: @Jorge please advise.
+
+**Example**
+
+.. code-block:: bash
+
+    certoraSolanaProver --build_script <path_to_build_script> --solana_inlining <path_to_inlining_file>
+
+.. _solana_summaries:
+--solana_summaries
+~~~~~~~~~~~~~~~~~~~
+
+**What does it do?**
+
+Provides the prover with a list of paths to summary files for Solana contracts. These files are parsed and used to prove properties.
+
+**When to use it?**
+
+TODO: @Jorge please advise.
+
+**Example**
+
+.. code-block:: bash
+
+    certoraSolanaProver --build_script <path_to_build_script> --solana_summaries <path_to_summaries_file>
+
+.. _cargo_features:
+--cargo_features
+~~~~~~~~~~~~~~~~~
+
+**What does it do?**
+
+Provides the prover with a whitespace-separated list of extra features passed to the build script.
+
+**When to use it?**
+
+TODO: @Jorge please advise.
+
+**Example**
+
+.. code-block:: bash
+
+    certoraSolanaProver --build_script <path_to_build_script> --cargo_features <feature_1> <feature_2>
+
+.. _msg:
+--msg
+~~~~~
+
+**What does it do?**
+
+Adds a description message to your run, similar to a commit message. This message appears in the title of the completion email.
+Note that you need to wrap your message in quotes if it contains spaces.
+
+**When to use it?**
+
+Adding a message makes it easier to track several runs. It is very useful if you are running many verifications simultaneously.
+It is also helpful to keep track of a single file verification status over time, so we recommend always providing an informative message.
+
+**Example**
+
+.. code-block:: bash
+
+    certoraSolanaProver --build_script <path_to_build_script> --msg 'Removed an assertion'
+
+.. _rule:
+--rule
+~~~~~~
+
+**What does it do?**
+
+Formally verifies one or more specified properties instead of the whole specification file. Can also verify an invariant.
+
+**When to use it?**
+
+This option saves a lot of run time. Use it whenever you care about only a
+specific subset of a specification's properties. The most common case is when
+This option saves runtime and is useful for verifying specific subsets of a specification. Common cases include testing new rules or investigating specific failures.
+you add a new rule to an existing specification. The other is when code changes
+cause a specific rule to fail; in the process of fixing the code, updating the
+rule, and understanding counterexamples, you likely want to verify only that
+specific rule.
+
+**Example**
+
+If `Bank.rs` includes the following:
+
+.. code-block:: text
+
+    invariant address_zero_cannot_become_an_account()
+    rule withdraw_succeeds()
+    rule withdraw_fails()
+
+To verify only `withdraw_succeeds`, run:
+
+.. code-block:: bash
+
+    certoraSolanaProver --build_script <path_to_build_script> --rule withdraw_succeeds
+
+To verify both `withdraw_succeeds` and `withdraw_fails`, run:
+
+.. code-block:: bash
+
+    certoraSolanaProver --build_script <path_to_build_script> --rule withdraw_succeeds withdraw_fails


docs/solana/options.rst
```diff
@@ -1,5 +1,201 @@
...
+~~~~~~~~~~~~~~~
+
+**What does it do?**
+
+Runs formal verification of specified properties while providing an automatic method to compile a Rust project.
+
+The build script must output the following:
+
+- `project_directory`: Path to the project root directory.
+- `sources`: List of files or directories used or imported in the program.
```
johspaeth, line 56: Can this list contain a regular expression? Or does the user need to list all files explicitly? 

```suggestion
- `sources`: List of source files or directories used or imported in the program. All files declared in this list will be uploaded as sources to cloud and displayed in the rule report. Source files are required, for instance, for the jump to source feature to work.
```

docs/solana/options.rst
```diff
@@ -1,5 +1,201 @@
...
+
+
+.. contents:: Overview
+   :depth: 2
+   :local:
+
+Modes of Operation
+------------------
+
+The Certora Solana Prover has two modes of operation. These modes are mutually exclusive - you cannot run the tool with more than one mode at a time.
```
johspaeth, line 43: What are the two modes of operations here? 

Are you referring to `--build_script` and `precompiled binary` here? That's not clear to me - state them explicitly?

docs/solana/options.rst
```diff
@@ -1,5 +1,201 @@
...
+    certoraSolanaProver --help
+
+Using Configuration (Conf) Files
+--------------------------------
+
+For larger projects, the command line for running the Certora Prover can become large and cumbersome. It is therefore recommended to use configuration files instead.
+
+These files are in [JSON5](https://json5.org/) format and use a `.conf` extension. They hold the parameters and options for the Prover.
+
+For more details, see `Conf File <https://docs.certora.com/en/latest/docs/prover/cli/conf-file-api.html#conf-files>`_
```
johspaeth, line 33:
```suggestion
For more details, see `Conf File <https://docs.certora.com/en/latest/docs/prover/cli/conf-file-api.html#conf-files>`.
```

docs/solana/options.rst
```diff
@@ -1,5 +1,201 @@
...
+
+**When to use it?**
+
+This option saves a lot of run time. Use it whenever you care about only a
+specific subset of a specification's properties. The most common case is when
+This option saves runtime and is useful for verifying specific subsets of a specification. Common cases include testing new rules or investigating specific failures.
+you add a new rule to an existing specification. The other is when code changes
+cause a specific rule to fail; in the process of fixing the code, updating the
+rule, and understanding counterexamples, you likely want to verify only that
+specific rule.
```
johspaeth, line 179: I think it's sufficient to just write:
```suggestion
This option saves a lot of run time as it only executes a specific rule. This option saves runtime and is useful for verifying specific subsets of a specification. Common cases include testing new rules or investigating specific failures.
```

docs/solana/options.rst
```diff
@@ -1,5 +1,201 @@
...
+
+    certoraSolanaProver --build_script <path_to_build_script> --cargo_features <feature_1> <feature_2>
+
+.. _msg:
+--msg
+~~~~~
+
+**What does it do?**
+
+Adds a description message to your run, similar to a commit message. This message appears in the title of the completion email.
```
johspaeth, line 149: The message is more important in the prover dashboard / rule report. I think we should rather state this here.

```suggestion
Adds a description message to your run, similar to a commit message. This message appears on the Prover dashboard and in the rule report.
```

docs/solana/options.rst
```diff
@@ -1,5 +1,201 @@
...
+--solana_inlining
+~~~~~~~~~~~~~~~~~~
+
+**What does it do?**
+
+Provides the prover with a list of paths to inlining files for Solana contracts. These files are parsed and used to prove properties.
+
+**When to use it?**
+
+TODO: @Jorge please advise.
```
johspaeth, line 99: @phreppo can you add a few words on the usage of it?

docs/solana/options.rst
```diff
@@ -1,5 +1,201 @@
...
+--solana_summaries
+~~~~~~~~~~~~~~~~~~~
+
+**What does it do?**
+
+Provides the prover with a list of paths to summary files for Solana contracts. These files are parsed and used to prove properties.
+
+**When to use it?**
+
+TODO: @Jorge please advise.
```
johspaeth, line 117: @phreppo can you add a few words on the usage of it?

docs/solana/options.rst
```diff
@@ -1,5 +1,201 @@
...
+--cargo_features
+~~~~~~~~~~~~~~~~~
+
+**What does it do?**
+
+Provides the prover with a whitespace-separated list of extra features passed to the build script.
+
+**When to use it?**
+
+TODO: @Jorge please advise.
```
johspaeth, line 135: @phreppo are you familiar with this, can you add a few words on the usage of it?

johspaeth (commented):
Mostly good - minor nits. Thanks.

@phreppo can you please add information about the missing parts where Yoav requested Jorge's support, I believe you should be familiar with the flags that miss explanations?

docs/solana/options.rst
```diff
@@ -1,5 +1,201 @@
...
+
+The Certora Solana Prover has two modes of operation. These modes are mutually exclusive - you cannot run the tool with more than one mode at a time.
+
+.. _build_script:
+--build_script
+~~~~~~~~~~~~~~~
+
+**What does it do?**
+
+Runs formal verification of specified properties while providing an automatic method to compile a Rust project.
```
phreppo, line 51: I might be wrong here, but I thought that the build script is the Python script that builds the project *without* running the formal verification.
Am I wrong?

docs/solana/options.rst
phreppo, line 135:
```suggestion
**What does it do?**

Provides the prover with a whitespace-separated list of extra `Cargo features <https://doc.rust-lang.org/cargo/reference/features.html>` passed to the build script.
These features are then passed to ``cargo`` to build the project.

**When to use it?**

Use it when there is a need to enable a specific `Cargo feature <https://doc.rust-lang.org/cargo/reference/features.html>` to compile the source code.
```

phreppo: Done

docs/solana/options.rst
phreppo: For iniling and summaries we need to decide what is best to do.
We can either have an ad-hoc section where we describe in detail how they work, or we can just say that these files use an internal representation.
In the second case we can give the users some default files and can be copy-pasted in a project and should work most of the times.
The problem with the second approach is that those files will have to be updated every now and then.

Maybe it is easier for the moment to just mention that these files use a Prover-specific format and provide them.
What do you think @johspaeth and @caballa?

docs/solana/options.rst
phreppo: See https://github.com/Certora/Documentation/pull/335/files#r1922635107

docs/solana/options.rst
yoav-el-certora: I was actually copy pasting from the Solidity version, but somehow it got additional lines which are not intended, great catch!

docs/solana/options.rst
yoav-el-certora: Same as below, took this text from the original Solidity version, but I do agree this makes much more sense.
Committing changes.

docs/solana/options.rst
yoav-el-certora: You are partly right, our current design flow is based on either `build_script` or `precompiled binary`.
This makes this attribute more like a run mode, which when running the `certoraSolanaProver` it will build the project and run formal verification.

Additionally, the build `build_script` can be any executable file, and not necessarily Python.

docs/solana/options.rst
phreppo: Note: after our discussion during the Solana meeting we decided not to write a separate ad-hoc section, but rather just mention that these files are needed.
We can then give examples for both inling and summaries that can be just downloaded and plugged in.
The ones that I am using during the setups are these: https://github.com/Certora/solana-project-template/tree/main/certora.
In particular, see https://github.com/Certora/solana-project-template/blob/main/certora/cvt_inlining_core.txt and https://github.com/Certora/solana-project-template/blob/main/certora/cvt_summaries_core.txt.

docs/solana/options.rst
```diff
@@ -1,5 +1,202 @@
...
+
+The build script must output the following:
+
+- `project_directory`: Path to the project root directory.
+- `sources`: List of source files or directories used or imported in the program. Source files should be relative to the `project_directory` with support of wildcards. All files declared in this list will be uploaded as sources to cloud and displayed in the rule report. Source files are required, for instance, for the jump to source feature to work.
+- `executables`: List of compiled binary files, which are the target of the Rust program.
+- `success`: Boolean flag indicating if the build phase passed successfully.
+
```
agurfinkel, line 60: @yoav-el-certora 
Would be useful to have a subsection with exact specifications for the build script.
What I missing from the above:
1. build script must support `--json` flag
2. build script should output nothing or json (depending on the flag)
3. build script should output 0 on success and 1 on failure
4. exact format of the json output
5. build script may optionally provide paths to files for stdout and stderr of the build
6. path to inline and summary files

docs/solana/options.rst
```diff
@@ -1,5 +1,202 @@
...
+**When to use it?**
+
+Use this mode to prove properties on source code while providing an automatic compilation method. This is especially useful during development when files are modified frequently.
+
+**Example**
+
+.. code-block:: bash
+
+    certoraSolanaProver --build_script <path_to_build_script>
```
agurfinkel, line 69: Here and elsewhere, there is not `-rule` flag given, so the prover will return an error saying that there is no rule to prove.

docs/solana/options.rst
agurfinkel: Note that these files are generated during the build process and must be provided by the build script, not by the user.

docs/solana/options.rst
```diff
@@ -1,5 +1,202 @@
...
+    certoraSolanaProver --build_script <path_to_build_script> --cargo_features <feature_1> <feature_2>
+
+.. _msg:
+--msg
+~~~~~
+
+**What does it do?**
+
+Adds a description message to your run, similar to a commit message. This message appears on the Prover dashboard and in the rule report.
+Note that you need to wrap your message in quotes if it contains spaces.
```
agurfinkel, line 152: I find the "note" confusing. `msg` takes one argument. If I want a string with spaces, I will pass it based on the shell that I'm using. Most often it is achieved by putting the string in quotes. Since the documentation is written for a command line user, they know what to do.

docs/solana/options.rst
```diff
@@ -1,5 +1,202 @@
...
+**When to use it?**
+
+Adding a message makes it easier to track several runs. It is very useful if you are running many verifications simultaneously.
+It is also helpful to keep track of a single file verification status over time, so we recommend always providing an informative message.
+
+**Example**
+
+.. code-block:: bash
+
+    certoraSolanaProver --build_script <path_to_build_script> --msg 'Removed an assertion'
```
agurfinkel, line 163: Perhaps we can do a better example. Right before, we say "informative message". A generic message like the one in the example is likely to not be helpful in a dashboard with many rules from many projects.

docs/solana/options.rst
```diff
@@ -1,5 +1,202 @@
...
+
+.. _rule:
+--rule
+~~~~~~
+
+**What does it do?**
+
+Formally verifies one or more specified properties instead of the whole specification file. Can also verify an invariant.
```
agurfinkel, line 171: there are no 'invariants' in Solana right now. Please remove.

docs/solana/options.rst
```diff
@@ -1,5 +1,202 @@
...
+
+**When to use it?**
+
+This option saves a lot of run time. Use it whenever you care about only a
```
agurfinkel, line 175: In Solana, there is no support for implicit rules. @caballa, can you please confirm?

In all my applications, I always have to provide at least one rule.

docs/solana/options.rst
```diff
@@ -1,5 +1,202 @@
...
+rule, and understanding counterexamples, you likely want to verify only that
+specific rule.
+
+**Example**
+
+If `Bank.rs` includes the following:
+
+.. code-block:: text
+
+    invariant address_zero_cannot_become_an_account()
```
agurfinkel, line 188: not a good example

docs/solana/options.rst
```diff
@@ -1,5 +1,202 @@
...
+This option saves a lot of run time. Use it whenever you care about only a
+specific subset of a specification's properties. The most common case is when
+you add a new rule to an existing specification. The other is when code changes
+cause a specific rule to fail; in the process of fixing the code, updating the
+rule, and understanding counterexamples, you likely want to verify only that
+specific rule.
+
+**Example**
+
+If `Bank.rs` includes the following:
```
agurfinkel, line 184: By convention, rust code files start with lower case letter.

docs/solana/options.rst
```diff
@@ -1,5 +1,202 @@
...
+
+.. code-block:: text
+
+    invariant address_zero_cannot_become_an_account()
+    rule withdraw_succeeds()
+    rule withdraw_fails()
```
agurfinkel, line 190: these seem to be examples from CVL. In CVLR they would be `rule_withdraw_succeeds` and the full name, inluding `rule_` should be used as a rule name.

docs/solana/options.rst
```diff
@@ -1,5 +1,202 @@
...
+
+To verify only `withdraw_succeeds`, run:
+
+.. code-block:: bash
+
+    certoraSolanaProver --build_script <path_to_build_script> --rule withdraw_succeeds
```
agurfinkel, line 196: should be `rule_withdraw_succeeds`

docs/solana/options.rst
```diff
@@ -1,5 +1,202 @@
...
+
+To verify both `withdraw_succeeds` and `withdraw_fails`, run:
+
+.. code-block:: bash
+
+    certoraSolanaProver --build_script <path_to_build_script> --rule withdraw_succeeds withdraw_fails
```
agurfinkel, line 202: names of rules should be fixed

docs/solana/options.rst
caballa: Yes, in solana you have to always provide explicitly all the rules. Also as you say we don't have invariants.

PR #336, opened 2025-01-20
DOC-431: Solana troubleshooting section

phreppo:
Jira ticket: https://certora.atlassian.net/browse/DOC-431

Comments:
* I changed the name from Solana Prover Troubleshooting just for consistency: the Troubleshooting section for Soroban is called simply Troubleshooting. Let me know if I have to change this.
* There are no possible problems with summarization and inlining using the `certoraSolanaProver` command, since the endpoint checks that summaries are inlining are present. For this reason, I did not include that section.
* For the JTS feature to work, the source files have to be uploaded. This is not done in `.conf` files, but rather in the `certora_build.py` script that we use to build Solana projects. I just specified that files have to be uploaded correctly, without saying how, since I am not sure if this is specified at all in any other section. If this is the case, I can link that section.


docs/solana/troubleshooting.rst
```diff
@@ -1,8 +1,36 @@
...
+``-C strip=none -C debuginfo=2``.
+Furthermore, the feature will not work if the source files are not correctly
+uploaded to the cloud.
+
+Prover Errors
+-------------
+
+The Solana Prover is currently under development, and some features are not
+supported yet.
+The most common source of errors from the Prover are stack-allocated arrays.
```
johspaeth, line 33: Is this producing an exception that the user should be aware of and look into the logs? Can we catch this exception and report it via `CVTAlertReporter` to the Global Notification tab. If not, can we explain the shape of the exception in textual here? I wouldn't know when I should start munging here.

docs/solana/troubleshooting.rst
```diff
@@ -1,8 +1,36 @@
...
+
+Jump to Source feature not working
+----------------------------------
+
+For the Jump to Source feature to work, the source code must be compiled with
+debug information enabled.
+This can be done by using the following compiler flags:
+``-C strip=none -C debuginfo=2``.
+Furthermore, the feature will not work if the source files are not correctly
+uploaded to the cloud.
```
johspaeth, line 26: > For the JTS feature to work, the source files have to be uploaded. This is not done in .conf files, but rather in the certora_build.py script that we use to build Solana projects. I just specified that files have to be uploaded correctly, without saying how, since I am not sure if this is specified at all in any other section. If this is the case, I can link that section.

This isn't properly described in https://github.com/Certora/Documentation/pull/335 either. I think we should describe this rather here then in the options (which doesn't talk much about JTS at all). 

I'd suggest we should link to a minimal example where it's working and then describe what the users needs to be aware of it to get it correct on their code base.

@yoav-el-certora do we have a minimal example that is public that we can link here and on which @phreppo can explain the potential issues?

johspaeth (commented):
Thanks for the draft.

Can we improve the description on the Stack vs Heap problem by adding a minimal example (i.e. adding the rust code that lead to an error and then how it needs to be fixed). Can we also help the user in understanding _when_ they ran into this issue?

docs/solana/troubleshooting.rst
phreppo: I updated the section to have: 1) a concrete example 2) a screenshot of the error 3) a possible solution.
The commit is 266847a1cd87b5a28b1f8f3d81645ec0ee8a6f4a.

docs/solana/troubleshooting.rst
phreppo: I am not sure that uploading the files to the cloud correctly belongs to troubleshooting or options.
It is more about "how to use the Solana workflow".
What would be the most appropriate section to specify that?
If there is no section that explains how to the workflow works (build files, summaries, etc...), maybe we should have one?

johspaeth: > It is more about "how to use the Solana workflow". 

Mh, right ... Maybe we should rather have a paragraph on "why can I not see any files?" and then in the section "How to use the Solana workflow" (Jorge is writing it) ensure that we state JTS only works when these files are present?

I leave the final decision to you here.

phreppo: I wrote a short paragraph about that in 2bf62e26fe44cf6a16ed14107e32639fd4c88e9a.
I just say that JTS not working can be due to files not correctly uploaded, and then I link the installation section.

johspaeth (approved):
LGTM. 

I think we should only make all files consistent to markdown. Wdyt?

PR #348, opened 2025-02-04
Fix unsupported --build_script flag

phreppo:


docs/solana/options.md
```diff
@@ -7,13 +7,13 @@ The ``certoraSolanaProver`` utility invokes the Rust compiler and then sends the
 The most commonly used command is:
 
 ```bash
-certoraSolanaProver --build_script <path_to_build_script>
+certoraSolanaProver <path_to_config\>.config
```
urikirsh, line 10:
```suggestion
certoraSolanaProver <path_to_config\>.conf
```
We use `.conf` in the example

urikirsh (changes requested):
Good, but we also have a section titled `--build_script` that starts from line 35, can you please remove it? Maybe add it to a different branch so that we don't lose it - we will want to merge it right after the fix

phreppo:
The branch is here: https://github.com/Certora/Documentation/tree/phreppo/solana-section-with-script

PR #349, opened 2025-02-05
Fixes to Solana documentation

phreppo:


docs/solana/options.md
```diff
@@ -29,8 +29,9 @@ For more details, see [Conf File](https://docs.certora.com/en/latest/docs/prover
 
 ## Modes of Operation
 
-The Certora Solana Prover has two modes of operation, using a predefined build script, and passing precompiled binary directly.
+The Certora Solana Prover has two modes of operation, using a build script, or passing a precompiled binary directly.
 These modes are mutually exclusive - you cannot run the tool with more than one mode at a time.
```
urikirsh, line 33: We also have a third option - running from a `.conf` file, like we mention in `SolanaExamples` (`certoraSolanaProver Default.conf`). This mode doesn't require using the `--rule` flag but requires a rules item in the `.conf` file

phreppo: I think that running from a configuration file is not really a mode of operation, as under the wood one of the two others is used.

urikirsh: I see we explain about conf files in `usage.md`, so it is indeed not needed here.

urikirsh (approved):
Thanks for all the changes

PR #352, opened 2025-02-11
Document sum and usum

naftali-g:
Jira ticket: TODO
Link to generated documentation: https://certora-certora-prover-documentation--352.com.readthedocs.build/en/352/docs/cvl/expr.html


docs/cvl/expr.md
```diff
@@ -242,6 +243,29 @@ approximation is {term}`sound`: it won't cause violations to be hidden.  See
 {ref}`grounding` for more detail.
 ```
 
+* The prover is capable of calculating the `sum` of ghost mappings over
+specified indices. For example, if we have a ghost mapping `ghost
+mapping(address => mapping(bytes32 => mathint)) myGhost`, and want to sum all
+the values of the ghost over all addresses for a given `bytes32` value `b`, one
+could write `sum address a. myGhost[a][b]`. This will return a `mathint`-typed
```
johspaeth, line 250:
```suggestion
can write `sum address a. myGhost[a][b]`. This returns a `mathint`-typed
```

docs/cvl/expr.md
```diff
@@ -242,6 +243,29 @@ approximation is {term}`sound`: it won't cause violations to be hidden.  See
...
+over it.
+
+* An extension of the summing logic is the unsigned sum which uses the `usum`
+keyword. It follows the same rules as the regular sum, but adds some extra
+logic to ensure the value of the sum is always larger than its parts.
+
+```{note}
+A `usum` can only be used on ghost mappings with unsigned (or `mathint`) value
+types. In the case of `mathint` an assertion is added on each write to the ghost
+that the written value is indeed unsigned.
```
johspaeth, line 267: Can we explain the difference between `usum` and `sum` better here? In particular, what's the difference of `sum` over an unsigned ghost mapping vs `usum` over the same mapping? @jhoenicke can you take a look and make a suggestion here?

johspaeth (changes requested):
I think we should clarify the difference between `usum` and `sum` and we can also link an example of the Example Repository for additional clarification.

docs/cvl/expr.md
jhoenicke: Maybe we can just add this

```
The keyword `usum` indicates that all entries are non-negative (unsigned). 
It can only be used on ghost mappings with unsigned (or `mathint`) value
types. In the case of `mathint` an assertion is added on each write to the ghost
that reports an error if the written value is negative.
When using this keyword, the solver will introduce the additional assumption
that the `usum` is larger than any value in the ghost mapping and that it is 
even larger than the sum of any finite subset of values.
This additional assumption is valid, because the other values in the ghost
mapping can make the total sum only larger. 
```

It's a bit repeating what was written above it, but maybe it's okay to repeat that here.

docs/cvl/expr.md
```diff
@@ -242,6 +243,36 @@ approximation is {term}`sound`: it won't cause violations to be hidden.  See
...
 ```
 
+* The prover is capable of calculating the `sum` of ghost mappings over
+specified indices. For example, if we have a ghost mapping `ghost
+mapping(address => mapping(bytes32 => mathint)) myGhost`, and want to sum all
+the values of the ghost over all addresses for a given `bytes32` value `b`, one
+can write `sum address a. myGhost[a][b]`. This returns a `mathint`-typed
+number that sums all known values of `myGhost` over the first index.
+The full syntax for this is `sum type1 t1, type2 t2, ... typeN tN.
+ghostName[...]`. See [here](https://github.com/Certora/Examples/blob/master/CVLByExample/Summarization/GhostSummary/GhostSums/README.md)
```
johspaeth, line 253: @yoav-el-certora fyi - we are hard-coding the link to the example on `master` in the docs here (even though it hasn't been merged). This breaks if the files are moved. Please ensure to keep the example in this location.

johspaeth (approved):
LGTM

PR #354, opened 2025-02-12
CERT-8241: revert in cvl

christiane-certora:
Jira ticket: TODO
Link to generated documentation: https://certora-certora-prover-documentation--354.com.readthedocs.build/en/354/docs/cvl/statements.html


docs/cvl/statements.md
```diff
@@ -345,6 +346,26 @@ assert lastReverted, "Expected revert when value exceeds limit";
 
 In this example, the `@withrevert` modifier is applied to the `Deposit` function call, which is expected to revert if the `value` exceeds the specified `limit`. The `assert` statement checks whether `lastReverted` is true, ensuring that the contract execution does revert as anticipated when the condition is violated. The error message in the `assert` provides additional context about the expectation.
 
+By default, this applies only for Solidity function calls. With the `--prover_args "-cvlFunctionRevert"` option, CVL functions can revert as well, and the `revert` statement becomes available in CVL. In this mode, CVL functions also set the `lastReverted` variable, just like Solidity functions and reverts of calls without `@withrevert` are propagated up to their callers. Only if no call in the chain had a `@withrevert` annotation do we assume no revert happened. A call with `@withrevert` stops propagation of reverts and will not make the calling function revert, but instead allows reading the `lastReverted` variable immediately after it to check whether the call reverted.
```
johspaeth, line 349:
```suggestion
By default, this applies only for Solidity function calls. With the `--prover_args "-cvlFunctionRevert"` option, CVL functions can revert as well, and the `revert` statement becomes available in CVL. In this mode, CVL functions also set the `lastReverted` variable, just like Solidity functions and reverts of calls without `@withrevert` are propagated up to their callers. Only if no call in the call stack had a `@withrevert` annotation do we assume no revert happened. A call with `@withrevert` stops propagation of reverts and will not make the calling function revert, but instead allows reading the `lastReverted` variable immediately after it to check whether the call reverted.
```

docs/cvl/statements.md
```diff
@@ -345,6 +346,26 @@ assert lastReverted, "Expected revert when value exceeds limit";
...
+    }
+}
+
+rule revertInCVL {
+    bool b;
+    cvlFunctionThatMayRevert@withrevert(b);
+    assert lastReverted <=> !b;
+}
+```
+- [Further examples](https://github.com/Certora/Examples/blob/ae2eca20d8e6caf378ff10cf8066ecfc45d3658d/CVLByExample/RevertKeyWord/example.spec)
```
johspaeth, line 367: Let's rather link to the example location in the repository assuming the PR was merged already. As you say, the link may not be available after merging.

johspaeth (commented):
Only minor nits. I'd like to ask @urikirsh for his thoughts though. 

* Is it clear what was changed? 
* Should be list that the flag `-cvlFunctionRevert` is only present in prover version `>7.25.0`? (Technically, it has been around before it but we only want to recommend using it starting that version).

docs/cvl/statements.md
christiane-certora: I did it like this because all the links in the page where such blob links, so I wasn't sure what's the usual process - do we just merge the Examples PR first and then update the link after?

docs/cvl/statements.md
```diff
@@ -345,6 +346,26 @@ assert lastReverted, "Expected revert when value exceeds limit";
 
 In this example, the `@withrevert` modifier is applied to the `Deposit` function call, which is expected to revert if the `value` exceeds the specified `limit`. The `assert` statement checks whether `lastReverted` is true, ensuring that the contract execution does revert as anticipated when the condition is violated. The error message in the `assert` provides additional context about the expectation.
 
+By default, this applies only for Solidity function calls. With the `--prover_args "-cvlFunctionRevert"` option, CVL functions can revert as well, and the `revert` statement becomes available in CVL. In this mode, CVL functions also set the `lastReverted` variable, just like Solidity functions and reverts of calls without `@withrevert` are propagated up to their callers. Only if no call in the call stack had a `@withrevert` annotation do we assume no revert happened. A call with `@withrevert` stops propagation of reverts and will not make the calling function revert, but instead allows reading the `lastReverted` variable immediately after it to check whether the call reverted.
```
urikirsh, line 349: If this option is user facing, it should have its own flag and not just a Prover arg. Then it would have better validation, better error messages, and appear in `--help`.
creating a flag from a prover arg is very easy, you can see an example by looking for the string `OPTIMISTIC_HASHING` in the file `scripts/CertoraProver/certoraContextAttributes.py`

urikirsh (commented):
If an option is exposed to users, it should be a flag, not a `prover_arg`. A flag is easier to use, has better validation and error messages, and appears in `--help`.

docs/cvl/statements.md
```diff
@@ -366,6 +366,10 @@ rule revertInCVL {
 ```
 - [Further examples](https://github.com/Certora/Examples/blob/ae2eca20d8e6caf378ff10cf8066ecfc45d3658d/CVLByExample/RevertKeyWord/example.spec)
 
+```{note}
+It is possible to enable the old behavior where CVL functions cannot revert and do not set the `lastReverted` variable with the `--prover_args "-cvlFunctionRevert false"` option. This is meant as a compatibility option where specs need to be adjusted to the new behavior and will be retired in a future release.
```
johspaeth, line 370:
```suggestion
It is possible to enable the behavior before version TODO where CVL functions cannot revert and do not set the `lastReverted` variable with the `--prover_args "-cvlFunctionRevert false"` option. This is meant as a compatibility option where specs need to be adjusted to the new behavior and will be retired in a future release.
```

johspaeth (approved):
Looks good to me.

Before merging this we need to remember to fill in the correct version. Leaving this comment as a reminder to ourselves.

docs/cvl/statements.md
```diff
@@ -345,6 +346,30 @@ assert lastReverted, "Expected revert when value exceeds limit";
 
 In this example, the `@withrevert` modifier is applied to the `Deposit` function call, which is expected to revert if the `value` exceeds the specified `limit`. The `assert` statement checks whether `lastReverted` is true, ensuring that the contract execution does revert as anticipated when the condition is violated. The error message in the `assert` provides additional context about the expectation.
 
+Since version TODO, this applies not only for Solidity function calls. CVL functions can now revert as well, and the `revert` statement is available in CVL. CVL functions also set the `lastReverted` variable, just like Solidity functions and reverts of calls without `@withrevert` are propagated up to their callers. Only if no call in the call stack had a `@withrevert` annotation do we assume no revert happened. A call with `@withrevert` stops propagation of reverts and will not make the calling function revert, but instead allows reading the `lastReverted` variable immediately after it to check whether the call reverted.
```
johspaeth, line 349:
```suggestion
Since version 8, this applies not only for Solidity function calls. CVL functions can now revert as well, and the `revert` statement is available in CVL. CVL functions also set the `lastReverted` variable, just like Solidity functions and reverts of calls without `@withrevert` are propagated up to their callers. Only if no call in the call stack had a `@withrevert` annotation do we assume no revert happened. A call with `@withrevert` stops propagation of reverts and will not make the calling function revert, but instead allows reading the `lastReverted` variable immediately after it to check whether the call reverted.
```

johspaeth (approved):
@urikirsh this PR is for version 8. I added a release8.0 label for it.

PR #372, opened 2025-03-27
removing lastHasThrown

urikirsh:
Removing the `lastHasThrown` keyword which has no effect in Solidity v0.8.0+ and is almost identical to `lastReverted` in older Solidity versions too.
Link to generated documentation: https://certora-certora-prover-documentation--372.com.readthedocs.build/en/372/docs/cvl/expr.html


jar-ben (approved):
lgtm

PR #373, opened 2025-03-31
Updating documentation for Solana to include multi-assert and sanity …

johspaeth:
This PR updates

1. the sanity documentation to match the current implementation
2. Add `multi_assert_check` and `rule_sanity` to documentation of commonly used flags.


docs/solana/options.md
```diff
@@ -180,3 +180,43 @@ It is also helpful to keep track of a single file verification status over time,
 ```bash
 certoraSolanaProver --build_script <path_to_build_script> --msg 'Removed an assertion' --rule <rule_name>
 ```
+
+### --rule_sanity
+
+**What does it do?**
+
+When used with `--rule_sanity: basic`, this flag executes a job in sanity mode which performs a vacuity check for a rule.
```
phreppo, line 188: Do we want to also add `--rule_sanity: advanced` here?

docs/solana/options.md
```diff
@@ -180,3 +180,43 @@ It is also helpful to keep track of a single file verification status over time,
...
+
+```bash
+certoraSolanaProver --build_script <path_to_build_script> --rule_sanity basic
+```
+
+
+## `--multi_assert_check`
+
+**What does it do?**
+This flags translates each assertion statement in a rule into a separate verification task. All preceding assertions are assumed to be true.
```
phreppo, line 204: tioli
```suggestion
This flags translates each assertion statement in a rule into a separate verification task. All preceding assertions are assumed to be true - they are hence translated into assume statements.
```

docs/solana/options.md
```diff
@@ -180,3 +180,43 @@ It is also helpful to keep track of a single file verification status over time,
...
+```bash
+certoraSolanaProver --build_script <path_to_build_script> --rule_sanity basic
+```
+
+
+## `--multi_assert_check`
+
+**What does it do?**
+This flags translates each assertion statement in a rule into a separate verification task. All preceding assertions are assumed to be true. 
+An example can be found [here](https://github.com/Certora/SolanaExamples/blob/main/cvlr_by_example/first_example/certora/conf/MultiAssertMode.conf).
```
phreppo, line 205: Should we point a specific commit rather than `main`? If we change the path of the file in `main,` the link brokes.

Another option is to give an example here.
This is not optimal, since if the conf file gets stale (no longer supported by the Prover), we would not observe that.
In the examples repo, the CI is executed daily.

phreppo (approved):
LGTM

docs/solana/options.md
johspaeth: I first want to further evaluate advanced mode internally, before we proceed documenting it. We may want to add other checks etc.

docs/solana/options.md
johspaeth: I'll update it to link to the commit (after the [other PR](https://github.com/Certora/SolanaExamples/pull/13) is resolved). I prefer _not_ having an explicit example here as of the reasons you state.

PR #380, opened 2025-04-15
Add require reason to statement grammar and explanation

christiane-certora:
To go with https://github.com/Certora/EVMVerifier/pull/7383 

Jira ticket: TODO
Link to generated documentation: https://certora-certora-prover-documentation--380.com.readthedocs.build/en/380/


docs/cvl/statements.md
```diff
@@ -21,7 +21,7 @@ block ::= statement { statement }
 
 statement ::= type id [ "=" expr ] ";"
 
-            | "require" expr ";"
+            | "require" expr [ "," string ] ";"
             | "static_require" expr ";"
```
jar-ben, line 25: Not related to this PR. `static_require` and `static_assert` are deprecated according [to our docs](https://docs.certora.com/en/latest/docs/cvl/cvl2/changes.html#static-assert-and-static-require). Can you check if we still support them in the prover? If not, we should remove them from the docs (in another Doc PR).

docs/cvl/statements.md
```diff
@@ -21,7 +21,7 @@ block ::= statement { statement }
 
 statement ::= type id [ "=" expr ] ";"
 
-            | "require" expr ";"
+            | "require" expr [ "," string ] ";"
```
jar-ben, line 24: What about the `require (expr, "string")` syntax, i.e. the one with `(...)`. Do we mention is somewhere in the docs? We should. Also for `assert` and `satisfy` if the prover supports that syntax.

docs/cvl/statements.md
```diff
@@ -71,9 +71,11 @@ a rule, while the `assert` statement is used to specify the expected behavior
 of contract functions.
 
 During verification, the Prover will ignore any {term}`model` that causes the
-`require` expressions to evaluate to false.  Unlike Solidity, the `require`
-statement does not contain a descriptive message, because the Prover will never
-consider an example where the `require` statement evaluates to `false`.
+`require` expressions to evaluate to false.  The require statement can contain 
+an optional descriptive message, and it is advised to specify it to document the
```
jar-ben, line 75: "to specify a message" - that does not sound like a good english to me, but I am not a native speaker. Also, the sentence is quite long and can be hard to read. You also use two words to refer to the same thing: "require expression" & "require statement", and possibly also "assumption".  

I think that we want to tell the readers that i] the require statement reduces the set of explored executions / models, ii] when used in-appropriately it can lead to unsoundness, i.e. excluding real violations of the property/rule, and iii] hence the user should carefully think about every `require` and a good explanation for it - the explanation can be added to the require statement.

docs/cvl/statements.md
christiane-certora: I don't find them in the parser/lexer, so probably they have indeed been removed. Will open a PR :+1:

docs/cvl/statements.md
christiane-certora: We have a note down in the explanation section that says:
> Unlike Solidity’s assert and require, the CVL syntax for assert and require does not require parentheses around the expression and message.

But we don't write anywhere explicitly that you can put them anyway. I just used the way we have specified it for assert and satisfy as guidance here, but I can add some `[(] ... [)]` around in the syntax section for all of them if you think it's better? I'm honestly not quite sure, because it might just make it look more cluttered and harder to read for something that you don't really need to know about and might just discover because it's intuitive to write like that. How do we handle documenting this kind of syntax varieties in general, do we have something else like that we can look at?

docs/cvl/statements.md
christiane-certora: I made a new attempt at this part, let me know if you like it better :+1:

docs/cvl/statements.md
jar-ben: You are right that adding `[(] ... [)]` would make the expression hard to read. So, let's just keep it as it is now.

docs/cvl/statements.md
jar-ben: I like the new wording, thanks!

jar-ben (approved):
lgtm. Do not merge it yet - we will merge it together with the prover release. Also, please add the "release" tag (you tagged it with "future"; if you check already closed PRs, you will see we use "release" for DOC PRs that should be merged during the prover main release).

PR #386, opened 2025-04-17
added --prover_version

urikirsh:
Add to the docs flags that appeared in the `--help` message that we want to keep exposed.

Link to generated documentation:


shoham-certora (approved):
Approved

PR #392, opened 2025-04-29
Add link to crates.io to CVLR documentation

johspaeth:
This PR adds a short description about where to find the CVLR packages.

### DIFF
--- a/docs/solana/speclanguage.md
+++ b/docs/solana/speclanguage.md
@@ -6,6 +6,10 @@ pre- and post-conditions for client code. Unlike CVL for Solidity, CVLR is
 embedded in Rust. It is compiled by the Rust compiler and has simple operational
 semantics.
 
+CVLR is deployed to [crates.io](https://crates.io/cvlr). The development version can be found at https://github.com/Certora/cvlr.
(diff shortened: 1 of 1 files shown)


docs/solana/speclanguage.md
```diff
@@ -6,6 +6,10 @@ pre- and post-conditions for client code. Unlike CVL for Solidity, CVLR is
 embedded in Rust. It is compiled by the Rust compiler and has simple operational
 semantics.
 
+CVLR is deployed to [crates.io](https://crates.io/cvlr). The development version can be found at https://github.com/Certora/cvlr.
+Please note that CVLR is independent from the Solana ecosystem, specific CVLR features for 
+Solana are maintained as part of the [cvlr-solana](https://crates.io/cvlr-solana) package.
```
johspaeth, line 11: Let me know if we should or shouldn't mention the solana package here. 

I think we should briefly state when to use which package and why - @1arie1 can you elaborate?

1arie1: The link is wrong. It should be https://crates.io/crates/cvlr-solana

1arie1 (approved):

PR #394, opened 2025-05-12
Document `--max_concurrent_rules`

chandrakananandi:


docs/prover/cli/options.md
```diff
@@ -1715,6 +1715,18 @@ This option disables the storage splitting optimization.
 --prover_args '-enableStorageSplitting false'
 ```
 
+(-maxConcurrentRules)=
+### `maxConcurrentRules`
+
+This allows setting the maximum number of concurrent rule evaluations. 
+Default value is the number of processors. This flag can sometimes mitigate out of memory problems.
```
urikirsh, line 1722: How? The user should set a higher or lower number? When should it be used?

docs/prover/cli/options.md
```diff
@@ -1231,6 +1231,34 @@ The second use is when the solvers can prove the property, they just need more t
...
+By default, it is set to the number of available CPU cores on the host machine, allowing optimal parallelization under typical conditions.
+Setting this value limits the number of rule evaluation tasks that can run in parallel.
+By capping the concurrency level, this attribute helps regulate system resource usage, particularly memory consumption, 
+and can prevent resource exhaustion.
+
+
+**When to use it?**
+
+Use this attribute when you are encountering out-of-memory errors, particularly when processing unusually large or complex rule sets.
+Lowering the concurrency can reduce peak memory usage by staggering evaluation workloads.
```
urikirsh, line 1254: What is the default concurrency level on the cloud on production?
Is it a good idea to write it down in our documentation? @H00N24

H00N24: I don't see any cloud-based modifications for this flag, so it should be the prover's default value.

H00N24: By default, we have 18 vCPUs available.

jar-ben (approved):

PR #406, opened 2025-05-28
Adds note regarding unsoundness of ecrecover axiom

shoham-certora:
Also replaced code block with code imported from the Examples repo.

Link to generated documentation: https://docs.certora.com/en/shoham-unsafe-ecrecover/docs/user-guide/patterns/safe-assum.html

### DIFF
--- a/docs/user-guide/patterns/safe-assum.md
+++ b/docs/user-guide/patterns/safe-assum.md
@@ -1,60 +1,22 @@
+```{role} cvl(code)
+:language: cvl
+```
+
 # Listing Safe Assumptions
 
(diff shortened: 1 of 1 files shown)


jar-ben (approved):
Just a remark - we have [recently introduced](https://docs.certora.com/en/latest/docs/cvl/statements.html#assert-and-require) the `require <expr><reason>` syntax where you can provide an explanation/justification for the require directly in the command (and we then have some tooling over that, like warnings in VSCode or notifications in the rule report). Perhaps it could be used in this example. 

But we also want to update more of our example to include the reasons (it will take some time since we have many examples though).

PR #424, opened 2025-07-28
Yoav/cli breaking changes

yoav-el-certora:
Naming convention:

Before requesting review:
 - Make sure CI is passing

Jira ticket: TODO


urikirsh (approved):
thanks for updating

docs/prover/cli/options.md
```diff
@@ -2322,6 +2322,43 @@ _Configuration file_
...
+```
+
+**What does it do?**
+This attribute controls the visibility of the report link generated by the CLI.
+To minimize the risk of unintentionally sharing a public URL, the CLI excludes anonymous access keys from the report link by default and returns only the private version.
+Enum values:
+    - private (default): Returns a report link without an anonymous access key.
+    - public: Returns a report link that includes an anonymous access key for public access.
+
+Note: In CI, the default behavior differs. The CLI will always output the public link to facilitate easy sharing among users.
```
urikirsh, line 2340:
```suggestion
```{Note} In CI, the default behavior differs. The CLI will always output the public link to facilitate easy sharing among users.
```
```

File: docs/prover/cli/options.md [hunk 2]
```diff
@@ -2322,6 +2322,43 @@ _Configuration file_
...
+
+**When to use it?**
+Use public when you are running the CLI locally and need to share the report externally.
```
urikirsh, line 2343:
```suggestion
Use `public` when you are running the CLI locally and need to share the report externally.
```

docs/prover/cli/options.md
```diff
@@ -2322,6 +2322,43 @@ _Configuration file_
...
+Use private when running jobs in CI environments to restrict access to the report.
```
urikirsh, line 2344:
```suggestion
Use `private` when running jobs in CI environments to restrict access to the report.
```

PR #433, opened 2025-08-22
Update documentation about dispatch lists

christiane-certora:
Update the grammar quite a bit (it was very outdated, still only had the deprecated syntax...), and add a section about dispatch list in the summaries.

Naming convention:

Before requesting review:
 - Make sure CI is passing

Jira ticket: TODO
Link to generated documentation: https://certora-certora-prover-documentation--433.com.readthedocs.build/en/433/docs/cvl/methods.html


docs/cvl/methods.md
```diff
@@ -722,6 +728,26 @@ Before this version, this may cause vacuous results.
...
+restrict which contract's implementations of the method to consider:
+```cvl
+methods {
+    function _.foo() external => DISPATCH(true)[ C._, D._ ];
+}
+```
+This will dispatch unresolved calls of a method `foo` to implementations in contracts
+`C` and `D`. 
+
+A `DISPATCH` summary is only useful on unresolved calls, so no policy should be specified.
```
johspaeth, line 750: I am not sure I understand, what do you mean by this?

docs/cvl/methods.md
```diff
@@ -722,6 +728,26 @@ Before this version, this may cause vacuous results.
...
+
+We can also use the `DISPATCH` list on {ref}`wildcard-methods-entries`, to
+restrict which contract's implementations of the method to consider:
+```cvl
+methods {
+    function _.foo() external => DISPATCH(true)[ C._, D._ ];
+}
+```
+This will dispatch unresolved calls of a method `foo` to implementations in contracts
+`C` and `D`.
```
johspaeth, line 748:
```suggestion
This will dispatch unresolved calls of a method `foo` to implementations in contracts
`C` and `D`.  Note, if `C` and `D` are the only contract implementing `foo` in the scene, then this is equivalent to using `DISPATCHER(true)`. 
```

docs/cvl/methods.md
```diff
@@ -722,6 +728,26 @@ Before this version, this may cause vacuous results.
...
+to a user-specified list of possible implementations.
+
+This is most commonly useful on calls where the method sighash is unresolved, see 
+{ref}`catch-unresolved-calls-entry` for a detailed example in such a case.
+
+We can also use the `DISPATCH` list on {ref}`wildcard-methods-entries`, to
+restrict which contract's implementations of the method to consider:
+```cvl
+methods {
+    function _.foo() external => DISPATCH(true)[ C._, D._ ];
```
johspaeth, line 744:
```suggestion
    function _.foo() external => DISPATCH(true)[ C.foo, D._ ];
```

...and then briefly explain that `foo` is implicit, so that we have both cases here.

johspaeth (commented):
Thanks! looks good, just a few comments.

docs/cvl/methods.md
christiane-certora: That you shouldn't put "ALL" or "DELETE" or something like that at the end. I copied the wording from the catch-unresolved-calls section where it says "Also, the visibility is always external, and no policy should be specified."

johspaeth (approved):

PR #438, opened 2025-09-11
ExecutingContract example

shellygr:
Naming convention:

Before requesting review:
 - Make sure CI is passing

Jira ticket: TODO


docs/cvl/methods.md
```diff
@@ -108,6 +108,33 @@ methods {
     function C.f(uint x) external returns(uint);
 }
 ```
+
+Similarly, `executingContract` can be used to refer to the caller that makes
```
shoham-certora, line 112: We already have an explanation of both `executingContract` and `calledContract`
in [expression-summaries documentation](https://docs.certora.com/en/latest/docs/cvl/methods.html#expression-summaries).  This addition would be better in that location.

Also we can use the examples from the `Examples` repo, e.g.
https://github.com/Certora/Examples/blob/651177a30d13e4c79034efa6c1cb5166601d065a/CVLByExample/TransientStorage/certora/specs/Mutexer.spec#L12)

shellygr: the example is for hooks so I prefer to separate the use cases

shoham-certora (approved):
Approved

PR #479, opened 2026-07-03
User guide: verification-idiom pages (unstructured storage, harnessing, vacuity, multi-call, relational math)

shellygr:
Adds pattern-oriented guidance for several recurring CVL spec-writing situations that were previously undocumented or only implicit in reference pages:

- **Unstructured storage** (`user-guide/patterns/unstructured-storage.md`): precomputing keccak-derived slot constants (EIP-1967, ERC-7201), binding them with `definition`, hooking them with `Sstore`/`Sload`, and mirroring them in ghosts.
- **Harnessing** (`prover/approx/harnessing.md`): getters for private/unstructured state, assembly `sload` getters, and decomposing a monolithic external function into thin external helper wrappers for step-by-step verification.
- **Vacuity root causes and repair** (`user-guide/patterns/vacuity.md`): why `NONDET` summaries on payable/side-effecting callees make callers revert on every path, and the repair order (sound summary → mock → `optimistic_fallback`) before ever filtering a method from parametric rules. Distinguishes the failure mode of unresolved `.call{value: ...}`/`send` (returns `false` on every path) from `transfer` (reverts on every path).
- **Multi-call properties** (`user-guide/patterns/multi-call-properties.md`): `storage s = lastStorage;` / `f(e, args) at s;` snapshot idioms for additivity (split vs. batch) and round-trip properties.
- **Satisfy witnesses** (`user-guide/satisfy.md` addition): companion `satisfy` rules proving implication antecedents reachable.
- **Relational math abstraction** (`user-guide/patterns/relational-math.md`): preferring monotonicity/bounds/round-trip properties over exact mulDiv formula mirroring, with an axiomatized `mulDivDownAbstract` example.
- **Reference-page reconciliation** (`cvl/expr.md`): documents `currentContract` as a valid storage-comparison basis, matching the `lastStorage[currentContract]` usage in the pattern pages.

CVL syntax in all examples was checked against the existing manual pages (hook grammar, `lastStorage`, `satisfy`, `requireInvariant`). New terms added to `spelling_wordlist.txt`. `user-guide/checking.md` cross-links the vacuity pattern page.

## Design revisions after architecture review

- Renamed `patterns/multicall.md` → `patterns/multi-call-properties.md` (page is about multi-call *properties*, not the Multicall contract pattern; URL becomes permanent at merge) — toctree and cross-links updated.
- Documented `currentContract` as a storage-comparison basis in `cvl/expr.md` so the reference page and the pattern pages' `lastStorage[currentContract]` snippets agree.
- Corrected the low-level-call sentence in `vacuity.md`: `transfer` reverts on failure rather than returning `false`, so its havoc symptom differs from `.call`/`send`.
- Added a one-line `envfree` note to each snippet that calls view functions without an `env` (`multi-call-properties.md`, `relational-math.md`, `satisfy.md`) so examples are copy-ready.
- Added a cross-link from `user-guide/checking.md` (vacuity section) to the new repair-sequence page.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


shellygr:
Delta since the previous version of this branch, for reviewers who already looked at it (new commits only — nothing was amended or force-pushed):

- Renamed `patterns/multicall.md` → `patterns/multi-call-properties.md` (the page covers multi-call *properties*, not the Multicall contract pattern); toctree entry in `patterns/index.md` and the cross-link in `relational-math.md` updated.
- `cvl/expr.md` now lists `currentContract` as a valid storage-comparison basis, so the reference page matches the `lastStorage[currentContract]` usage in the pattern pages.
- Fixed the low-level-call sentence in `vacuity.md`: `transfer` reverts on failure (it does not return `false`), so its havoc-induced vacuity symptom differs from `.call{value: ...}`/`send`.
- Added one-line `envfree` notes to snippets calling view functions without an `env` (`multi-call-properties.md`, `relational-math.md`, `satisfy.md`).
- `user-guide/checking.md` (vacuity section) now cross-links the repair-sequence pattern page.

PR description updated to match.

